PR #2, opened 2018-11-20
Plotting (skipped correlation)

LegrandNico:


raphaelvallat:
Thanks @LegrandNico for the pull request!

- I just looked at the tests (Travis and AppVeyor) and they are both failing because there is a missing comma in the setup.py between matplotlib and seaborn, should be:

```
'pandas>=0.23',
'matplotlib>=3.0.2',
'seaborn>=0.9.0'
```

- Also, I am wondering if the function should actually return either the figure (fig), axes (ax1, ax2, ax3), or both of them? That way the users would be able to make some final adjustements to the plot if they want to. 

- make sure to add the following line in the file pingouin/init.py:

```
from .plotting import *
```

and in the beginning of the plotting.py file:

```
__all__ = ["plot_skipped"]
```

- should we rename the function `plot_skipped_corr` to avoid any confusion?

- Ultimately we will also have to add some tests (i.e. create a new file tests/test_plotting.py), I think we will have to dig into the Seaborn code to understand how the tests are done.

Thanks again!
Raphael

LegrandNico:
The tests, for now, consist of simple figure generation with different kind of arguments. An alternative approach could be to compare the final plots with a pre-generated image using @pytest.mark.mpl_image_compare .

raphaelvallat:
@LegrandNico This looks great! Thank you for adding the tests!

Some minor final comments

- Please remove the blank line between the `"""` and `from pingouin.correlation import skipped` in the main function (line 27), just to be consistent with the other Pingouin functions.

- Please rename the test function to `def test_plot_skipped_corr(self)`

- Please remove the useless imports in the test_plotting.py file:

```
import pytest
import pandas as pd
import matplotlib.pyplot as plt
import seaborn as sns
```
- To simplify the output, I think we can just return the Figure `fig`, and if the users want the axes they can simply use `fig.axes`.

- Could you add the Returns in the documentation:
```
Returns
--------
fig : matplotlib Figure instance
    Matplotlib Figure. To get the individual axes, use fig.axes.
```

- I think we should remove the line `sns.set(style='ticks', context='notebook', font_scale=1.2)` from the function and instead let the user choose its own Seaborn preference. Or perhaps we can move it at the beginning of the file right after the import. That way the style/context will be set by default when Pingouin is imported but at least the user can change if afterwards. What do you think is best?

Thank you again! Such a pleasure to work on Pingouin's very first pull request!
Raphael

LegrandNico:
I moved the seaborn preference at the beginning of the file. While I completely agree that it is important to provide a hackable plot, it might also be nice to have homogeneous graphics compared to those we will possibly show in the documentation, especially if we add other plotting function in the future. So maybe it is preferable to declare these preferences the way you proposed, so the user can decide. Still, the graphical parameters I chose can be debated.

PR #50, opened 2019-07-18
Added TOST equivalence testing

Leicas:
I implemented a Two-One sided Test method for parametric and non-parametric.
The parametric implementation is tested against R equivalence package, I could not find any implementation for non-parametric to compare, but I followed the same logic of implementation and tested on obvious cases.

### DIFF
--- a/pingouin/equivalence.py
+++ b/pingouin/equivalence.py
@@ -1,0 +1,86 @@
+# Author: Antoine Weill--Duflos <antoine@weill-duflos.fr>
+# Date: July 2019
+import numpy as np
+import pandas as pd
+from itertools import combinations, product
+from pingouin.parametric import anova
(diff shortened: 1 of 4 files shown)


raphaelvallat:
Hi @Leicas ! Thank you so much for the PR, this looks great!

A few thoughts: 

1) Could you please submit the PR on the develop branch rather than the master? You will find that I've already added the `tail` argument to `pg.wilcoxon` and `pg.mwu` on the develop branch so there's no need to change this file.

2) Could you also please remove "noninf" from `__all__`?

3) Could you make sure that your code follows the flake8 guidelines? See https://pingouin-stats.org/contributing.html

4) Could you add some examples in the docstring of the `tost` function?

Thank you so much for your help!

PR #83, opened 2020-02-07
 Import ABC from collections.abc instead of collections for Python 3.9 compatibility.

tirkarthi:
*  Import ABC from collections.abc instead of collections for Python 3.9 compatibility.
* Fix deprecation warning due to invalid escape sequences.


tirkarthi:
Okay, I noticed it being vendored. The deprecation warning was fixed upstream and I copied the same code here to fix the warning in case tabulate will used without being updated.

PR #96, opened 2020-05-21
Update partial_corr to assert that x/y are independent from covar

jnecus:
During my scripts I was entering x/y/covar as the same string in some cases. This led to a strange error downstream which took a little time to trace back to the casue. So I have included these checks as assertions. There may be a better way to do this so that this condition is checked for all functions which require inputs to be independent..Let me know if you think so.


raphaelvallat:
That's great, merging now, thanks for the PR!

Issue #87, opened 2020-04-01
Assertion failure (-1 <= r <= 1) in power.py when running rm_corr


willprice:
Hi there,
Thanks for providing this awesome library. I picked it up to run an `rm_corr` test on some data but I'm having issues.

I have 1000 individuals for each of whom I have 8 paired measurements. I have uploaded a pickled dataframe that causes the issue to GDrive (https://drive.google.com/open?id=1gkvGKBth2J8EhmhblsWRdK-eCDlk8zJk).

Here's a snapshot of what the dataframe looks like:
![image](https://user-images.githubusercontent.com/966205/78142702-2add9c80-7425-11ea-8759-6f6a4046f44c.png)

You can reproduce the issue by running:

```python
df = pd.read_pickle('bad_df.pkl')
rm_corr(data=df, x='m2', y='m1', subject='id')
```

which yields the error

```
AssertionError                            Traceback (most recent call last)
<ipython-input-156-86b16d57ef4b> in <module>
----> 1 rm_corr(data=df, x='m2', y='m1', subject='id')

.../lib/python3.7/site-packages/pingouin/correlation.py in rm_corr(data, x, y, subject, tail)
    981     pval = pval * 0.5 if tail == 'one-sided' else pval
    982     ci = compute_esci(stat=rm, nx=n, eftype='pearson').tolist()
--> 983     pwr = power_corr(r=rm, n=n, tail=tail)
    984     # Convert to Dataframe
    985     stats = pd.DataFrame({"r": round(rm, 3), "dof": int(dof),

.../lib/python3.7/site-packages/pingouin/power.py in power_corr(r, n, power, alpha, tail)
    902     # Safety checks
    903     if r is not None:
--> 904         assert -1 <= r <= 1
    905         r = abs(r)
    906     if alpha is not None:

```

If I drop into a debugger I can see that `r = 1.0000184067367273`, perhaps this is due to numerical stability issues?

raphaelvallat:
Hi @willprice! Fascinating. Thanks for opening this issue and providing all the data and code required to replicate it. The issue is indeed caused by the fact that the ANOVA/ANCOVA functions that are called internally by rm_corr apply a rounding to the third decimals on the sums of squares. In your case, `m1` and `m2` are very small, which leads to an error. A quick fix for now is to do:

```
df[['m1', 'm2']] *= 100
rm_corr(data=df, x='m2', y='m1', subject='id')
```

which gives the following:

```
[...]

PR #101, opened 2020-06-15
adding option to return stats as float, rather than str

Spaak:
This PR is a *work in progress*.

I'm not sure whether there is a general guideline about returning statistics as formatted strings, rather than as float, but I find it very useful to use the raw float values, rather than get strings. This PR adds this functionality, at least to the `bayesian` module (not sure what the situation is in other modules).

This PR leaves the default behaviour of returning strings untouched. It could be argued that returning the raw float values should be default, and formatting left to the user in general (or some utility function perhaps). I'm interested to hear thoughts on this as well.

If you think in principle this PR is a good idea, I'll update and add proper documentation etc.

### DIFF
--- a/pingouin/bayesian.py
+++ b/pingouin/bayesian.py
@@ -17,7 +17,8 @@ def _format_bf(bf, precision=3, trim='0'):
     return out
 
 
-def bayesfactor_ttest(t, nx, ny=None, paired=False, tail='two-sided', r=.707):
+def bayesfactor_ttest(t, nx, ny=None, paired=False, tail='two-sided', r=.707,
+    return_as_str=True):
     """
     Bayes Factor of a T-test.
 
@@ -121,6 +122,8 @@ def bayesfactor_ttest(t, nx, ny=None, paired=False, tail='two-sided', r=.707):
     >>> print("BF10-greater: %s | BF10-less: %s" % (bf_greater, bf_less))
     BF10-greater: 0.029 | BF10-less: 34.369
     """
+    _format_fun = _format_bf if return_as_str else lambda x: x
+
     # Check tails
     possible_tails = ['two-sided', 'one-sided', 'greater', 'less']
     assert tail in possible_tails, 'Invalid tail argument.'
@@ -156,10 +159,11 @@ def bayesfactor_ttest(t, nx, ny=None, paired=False, tail='two-sided', r=.707):
     if ((tail == 'greater' and t < 0) or (tail == 'less' and t > 0)) and bf10 > 1:  # noqa
         bf10 = 1 / bf10
 
-    return _format_bf(bf10)
+    return _format_fun(bf10)
 
 
-def bayesfactor_pearson(r, n, tail='two-sided', method='ly', kappa=1.):
+def bayesfactor_pearson(r, n, tail='two-sided', method='ly', kappa=1.,
+    return_as_str=True):
     """
     Bayes Factor of a Pearson correlation.
 
@@ -274,6 +278,8 @@ def bayesfactor_pearson(r, n, tail='two-sided', method='ly', kappa=1.):
     >>> print("BF: %s" % bayesfactor_pearson(r, n, tail='one-sided'))
     BF: 21.185
     """
+    _format_fun = _format_bf if return_as_str else lambda x: x
+
     from scipy.special import gamma, betaln, hyp2f1
     assert method.lower() in ['ly', 'wetzels'], 'Method not recognized.'
     assert tail.lower() in ['two-sided', 'one-sided', 'greater', 'less',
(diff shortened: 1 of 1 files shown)


raphaelvallat:
Hi @Spaak ! Thanks for opening the issue. 

BFs were returned as raw float in the first versions of Pingouin. I disabled this behavior at some point because very small or very large numbers in the BF10 columns were changing the display format of ALL float columns in the output dataframe, which was very annoying (i.e. you don't need a scientific notation when all your t-values or effect sizes are within 0 and 5...). So even though I also prefer to report the raw float values instead of a string, I'd like to make sure that it does not affect the float format of other columns in the dataframe. Does that make sense? 

Curious to hear other users' thoughts so please feel free to chime in.

Thanks
Raphael

Spaak:
Hey, sorry for the slow reply. I did a bit of digging, and it turns out that Pandas has a [styling API](https://pandas.pydata.org/pandas-docs/stable/user_guide/style.html). Unfortunately, this only applies to HTML renderings of the dataframe, so is useless for general (console-based) number formatting per column.

I also saw in some other places in the code that Pingouin by default returns rounded values. In general I think rounding is good for the useability of the resulting dataframes (again for number format readibility reasons), but perhaps it would be good to have this as an option? I don't think there is right now a general framework for setting Pingouin options in place, right? I'd be happy to do some work to make this happen. I could see both rounding and stringifying as useful options to be able to switch on/off.

raphaelvallat:
Hi @Spaak,

Re: rounding. We had a similar conversation/questioning in a previous issue (https://github.com/raphaelvallat/pingouin/issues/87), and the conclusion was that Pingouin should avoid rounding and return the raw values whenever possible. I like the idea of a Pandas-based display formatting, i.e. you can still get the raw values with full precision but Pandas would only display X decimals. The issue however is that it would have to be different for different columns. For example, you typically don't want to have 5 digits for a F-value when running an ANOVA but you want the max precision for the p-values. So I'm not sure how we should create global display that would take that into account.

Happy to hear your thoughts,

Thanks
Raphael

Spaak:
> I like the idea of a Pandas-based display formatting

Yeah, I liked it also, but unfortunately this only works for HTML-based rendering, not for console work. It might still be worth pursuing, though.

> The issue however is that it would have to be different for different columns. For example, you typically don't want to have 5 digits for a F-value when running an ANOVA but you want the max precision for the p-values. So I'm not sure how we should create global display that would take that into account.

How about the following: Every function prepares a dataframe without any rounding, and then passes it to some general `postprocess_dataframe` function or so before returning it, which optionally applies rounding (and optionally also adds the HTML Pandas styles). Which rounding to apply should be governed by customizable options, e.g. `pingouin.set_option('round.default', 4)` for 4 decimals default; `pingouin.set_option('round.column.CI95%', 2)` for specific columns, etc. Specific option sets could be enabled (as e.g. seaborn does for plot styles) with `pingouin.set_round_options('compact')` or `pingouin.set_round_options(None)` etc., providing sensible defaults for all columns.

I'm on holiday at the moment, but if you agree this is a good idea, I'd be happy to implement it mid-August (will be back 10 Aug).

Cheers,
Eelke

JoElfner:
If I understand it correctly and this is about the displaystyle of resulting dataframes, then there is a way to set it independently of HTML formatting:
Either set the formatter explicitly:
`pd.set_option('display.float_format', lambda x: '{0:.3f}'.format(x))`
or set the precision, leaving the resulting format (scientific/not-scientific) decision to the module:
`pd.set_option('display.precision', 3)`
Also for numpy:
`np.set_printoptions(precision=3)`

Of course precision or format style can be adjusted depending on the data range.

Spaak:
@JoElfner Yes, pandas has the `display.float_format` option, but the drawback here is that this applies to the display of all numeric values. For the display of stats, we typically want different formats (or at least different decimal precisions) for different columns.

Meanwhile I started work on a PR implementing the `postprocess_dataframe` idea, so should be able to submit within a week or so.

raphaelvallat:
Hi @Spaak!

Now that the other PR is merged, should we add a default option to Pingouin to format the BayesFactor as an (exponential) float instead of a string?

Spaak:
> Now that the other PR is merged, should we add a default option to Pingouin to format the BayesFactor as an (exponential) float instead of a string?

Well yes, *however* I only later realized that the issue with the BF formatting is not so much that of rounding, but is a matter of order-of-magnitude. I don't think this is solvable while keeping BFs as (optionally rounded) floats (as opposed to strings). So, in other words, my other PR #120 does not solve the issue with BF formatting at all 🤦‍♂️.

I think, however, that the general mechanism I introduced in #120 is applicable: we could have a custom post-processing "rounding" function which optionally formats a cell/column/row in a particular way. This one we could set to default to the current behaviour (strings), yet then it would be easily customizable if people prefer floats.

I have a slight preference for exposing the custom formatters through the existing rounding options (as rounding and custom formatters would be mutually exclusive), like e.g. `pingouin.options['round.column.BF10'] = lambda x: np.format_float_scientific(x, precision=3, trim='0')`. Then the rounding code will simply check if the specified option is a callable; if so, apply it; if integer, apply rounding.

What do you think @raphaelvallat? I could easily file another PR to implement this if you agree it's a good idea.

raphaelvallat:
That's a great idea @Spaak!  And I think the default should be:

```python
from pingouin.bayesian import _format_bf
pingouin.options['round.column.BF10'] = lambda x: _format_bf(x, precision=3, trim='0')  # Careful, returns a str
```

In other words, we don't change the current behavior of Pingouin but we link it to your newly-added options system.

Thanks!

PR #113, opened 2020-08-03
Add processed arrays to linear_regression output

kncrabtree:
When as_dataframe=False, add arrays containing processed
regression data X, Xw, y, yw, and the model prediction pred to
the output dictionary.


raphaelvallat:
Hi @kncrabtree,

Thanks for the PR. Looks great, just one comment: to limit memory usage, should we only return Xw and Yw if weights are not None? In other words for standard OLS we only return X and y.

kncrabtree:
Good call; I've made that change.

PR #120, opened 2020-08-24
implement configurable rounding options

Spaak:
This PR adds a `pingouin.options` interface (via a new submodule `config`), and adds support for configurable rounding in `utils.postprocess_dataframe`. This function is now applied at the end of all functions that return their results as a pandas DataFrame.

See also #101, where this idea was discussed.


raphaelvallat:
Wow, this looks absolutely fantastic @Spaak !! Thank you so much for all the hard work! I cannot wait to try it. Just a few comments:

1) Could you submit the PR on the develop branch (working branch), and not the master (release-only branch)?

2) Could you please add an explanation of your changes to the _docs/changelog.rst_ file, with a quick example of how you can now use pingouin.options to apply a global or column-specific rounding? We can also add the example to the README.rst and docs/index.rst.

3) Please add your name to the list of contributors in the README.rst and docs/index.rst files!

Thanks again!

raphaelvallat:
Two more things we need to be careful of:

1) Are the hidden attributes preserved after rounding? For instance, the degrres of freedom in pingouin.linear_regression:

```
if as_dataframe:
  stats = pd.DataFrame(stats)
  stats.df_model_ = df_model
  stats.df_resid_ = df_resid
```

2) Depending on the function, the p-value column is sometimes called `pval`, `p-val`, `p-unc` or `p-corr`. Something I've been wanting to do for a long time is to homogenize the column names and remove the dashes, i.e. keeping only `pval`, `punc`, `pcorr` (or even just `pval` and `pcorr`).

Spaak:
> Wow, this looks absolutely fantastic @Spaak !! Thank you so much for all the hard work!

You're very welcome :)

> 1. Could you submit the PR on the develop branch (working branch), and not the master (release-only branch)?

Done.

> 2. Could you please add an explanation of your changes to the _docs/changelog.rst_ file, with a quick example of how you can now use pingouin.options to apply a global or column-specific rounding?

Done.

> We can also add the example to the README.rst and docs/index.rst.

I'm not sure exactly whether and how to include it. The examples given in README.rst and docs/index.rst seem very much about basic functionality to me (i.e., specific tests), I feel like the config/rounding mechanism doesn't really belong there (as people might only look for it after having used pingouin for some time already). If you disagree, let me know! (And then also some pointers on how to easily get those nice markdown tables would be welcome.)

> 3. Please add your name to the list of contributors in the README.rst and docs/index.rst files!

Done! Thanks.

> Are the hidden attributes preserved after rounding? For instance, the degrres of freedom in pingouin.linear_regression:
> 
> ```
> if as_dataframe:
>   stats = pd.DataFrame(stats)
>   stats.df_model_ = df_model
>   stats.df_resid_ = df_resid
> ```

I think this example from `linear_regression` was the only place I encountered such hidden attributes. There, I made sure to call `postprocess_dataframe` *before* the hidden attributes were added.

> Depending on the function, the p-value column is sometimes called `pval`, `p-val`, `p-unc` or `p-corr`. Something I've been wanting to do for a long time is to homogenize the column names and remove the dashes, i.e. keeping only `pval`, `punc`, `pcorr` (or even just `pval` and `pcorr`).

That makes sense indeed, but I'm inclined to say it's beyond the scope of this PR? (But I agree that it would be good, especially because then `pingouin.options['round.column.pval']` is respected everywhere!) Happy to do it in another PR.

raphaelvallat:
Hi @Spaak!

Sounds great on all fronts! Let's not change the readme and index.rst files for now, and similarly let's change the p-value column name in another PR. All the tests have passed so I'll merge the PR right away!

PS: To create RST (these are not markdown files) tables from a dataframe you can use the [pingouin.print_table](https://pingouin-stats.org/generated/pingouin.print_table.html#pingouin.print_table) function!

Thanks again!
Raphael

PR #124, opened 2020-08-31
pairwise_corr doc updated and extended functionality to automatically control for all remaining covariates

### LINKED ISSUE #123: pairwise_corr: unclear doc
Hi @raphaelvallat 

in my opinion, the doc of pairwise_corr is a bit unclear about the partial correlation. Imho it should be clarified, which correlation function (in this case `pg.corr`) is used, if `covar=None` and that partial corr. is only calculated, if `covar is not None`.

Furthermore, I'd like to see pairwise_corr to support the case that `columns=None` and that all columns are specified as covars, f.i. implicitly by setting `covar='all'` or explicitly by allowing to set `covar=data.columns`. This should calculate the `partial_corr` for all combinations of `x` and `y` while automatically excluding `x` and `y` from the covars.

If you think this could be useful, I'll start implementing it.

JoElfner:
I updated doc of `pairwise_corr` to reflect more clearly that a partial corr is only calculated, if covariates are specified.

Furthermore I implemented that combinations of columns, which are also present in covars, will not be dropped. Instead the `x` and `y` columns will be removed from `covar` only for this specific combination. This allows for automatically controlling for all other columns in the dataframe while calculating the pairwise **partial** correlation for all combinations.

If this means that there are not covars for a combination, an empty list will be returned for `stats['covar']` and via [`partial_corr` line 744](https://github.com/raphaelvallat/pingouin/blob/08560bb978949e97653b203f7374e90b022b88fb/pingouin/correlation.py#L744) a standard non-partial corr will be returned. Is this clear enough for the user or should "empty" covar rows be dropped instead?

Closes #123 

Some basic examples of the funcitonality (tests are included in the respective test file)
```
import numpy as np
import pandas as pd

import pingouin as pg

df = pd.DataFrame(
    data=np.random.rand(200, 4), columns=['a', 'b', 'c', 'd'])
df['e'] = df['a']**2
df['f'] = df['a'] * df['b']
df['g'] = df['c'] + np.random.randint(0, 5, 200)
df['h'] = np.sqrt(df['d'] + 5) - df['b']**3

# Get pairwise corr for all available covars per combination, also try nan policy
pwc = pg.pairwise_corr(data=df, covar=df.columns)
pwc_lwnan = pg.pairwise_corr(data=df, covar=df.columns, nan_policy='listwise')
# assert that both are equal
assert (
    pwc.drop(columns=['CI95%']) == pwc_lwnan.drop(columns=['CI95%'])
).all().all()

# Now the same with a MultiIndex:
mi = pd.MultiIndex.from_product((
    ('a', 'b', 'c', 'd'), ('u', 'v'),
))
df_mi = df.copy()
df_mi.columns = mi

pwc_mi = pg.pairwise_corr(data=df_mi, covar=df_mi.columns)
# Assert that the results are equal to non-multiindex tests
assert (
    pwc_mi.drop(
        columns=['X', 'Y', 'covar', 'CI95%']) == pwc.drop(
            columns=['X', 'Y', 'covar', 'CI95%'])
).all().all()
# and compare to nan policy listwise
pwc_mi_lwnan = pg.pairwise_corr(
    data=df_mi, covar=df_mi.columns, nan_policy='listwise'
)
assert (
    pwc_mi.drop(
        columns=['X', 'Y', 'covar', 'CI95%']) == pwc.drop(
            columns=['X', 'Y', 'covar', 'CI95%'])
).all().all()
```

And to show what I mean with "empty covars" (taken from test_pairwise.py):
```
from pingouin import pairwise_corr, read_dataset

data = read_dataset('pairwise_corr').iloc[:, 1:]
n = data.shape[0]
data['Age'] = np.random.randint(18, 65, n)
data['IQ'] = np.random.normal(105, 1, n)
data['One'] = 1
data['Gender'] = np.repeat(['M', 'F'], int(n / 2))

# see row 4 for empty covar
pwc_empty = pairwise_corr(data, columns='Neuroticism', covar='Age')
print(pwc_empty.covar)
# or fully empty covar
pwc_fully_empty = pairwise_corr(data, columns=['Neuroticism', 'Age'], covar='Age')
```


pingouin/tests/test_pairwise.py
```diff
@@ -495,8 +495,16 @@ def test_pairwise_corr(self):
         pairwise_corr(data, covar=['Age', 'Neuroticism'])
         with pytest.raises(AssertionError):
             pairwise_corr(data, covar=['Age', 'Gender'])
-        with pytest.raises(ValueError):
-            pairwise_corr(data, columns=['Neuroticism', 'Age'], covar='Age')
+        # with pytest.raises(ValueError):
+        #     pairwise_corr(data, columns=['Neuroticism', 'Age'], covar='Age')
```
JoElfner, line 499: This test was meant to fail, but now non-partial `corr` is calculated instead, returning an empty list for covar. What do you think about this?

JoElfner (commented):
Please specifically look at the commented test.

pingouin/pairwise.py
```diff
@@ -1132,12 +1132,12 @@ def traverse(o, tree_types=(list, tuple)):
...
-            raise ValueError("No column combination found. Please make sure "
-                             "that the specified columns and covar exist in "
-                             "the dataframe, are numeric, and contain at "
-                             "least two unique values.")
+        # # Assert that there is at least one combination
+        # if stats.shape[0] == 0:
+        #     raise ValueError("No column combination found. Please make sure "
+        #                      "that the specified columns and covar exist in "
+        #                      "the dataframe, are numeric, and contain at "
+        #                      "least two unique values.")
```
JoElfner, line 1140: Is this still required? This just seems to reduce codecov, while line 1104 already checks for zero-length data.

pingouin/tests/test_pairwise.py
```diff
@@ -539,6 +547,7 @@ def test_pairwise_corr(self):
         pairwise_corr(data, covar=[('Psycho', 'Anxiety')])
         pairwise_corr(data, columns=[('Behavior', 'Rating')],
                       covar=[('Psycho', 'Anxiety')])
+        pairwise_corr(data, covar=data.columns)
```
raphaelvallat, line 550: @JoElfner Can we add a test against the pingouin.pcorr() function? Just want to make sure that the r values we get are consistent with the latter.

raphaelvallat:
Hi @JoElfner!

Thanks so much for the PR, I just had a quick look and it looks absolutely great! I've added one comment re: unit testing against the pingouin.pcorr function. I'll have a more detailed look at the code within the next days, but before that, can you please:

1) Add the change to the docs/changelog.rst file
2) Add your name to the contributor list in the docs/index.rst file as well as readme.rst file
3) Submit the PR to the develop branch (working branch) rather than the master (for stable release only)

Thanks again!
Raphael

pingouin/tests/test_pairwise.py
JoElfner: Sure! done

JoElfner:
Hi @raphaelvallat,
also many thanks from my side!
I added testing against `pcorr` and update the docs and readme. Thanks for including me in your list of contributors!
Sorry for taking the wrong branch. While changing the branch to develop, I got some quite strange and unexpected conflicts. But here in the PR everything looks fine. Could you please double check for any unexpected regressions?

pingouin/pairwise.py
raphaelvallat: I think this was just to handle the case where you end up with an empty dataframe after `stats = stats[~stats[['X', 'Y']].isin(covar).any(1)]`, but if you find that it doesn't reduce coverage to remove it with your changes, then please feel free to remove it.

pingouin/tests/test_pairwise.py
raphaelvallat: The new behavior seems ok to me so ok to remove this test!

JoElfner:
Thanks!
I had already flaked everything before adding the additional tests. But when adding the tests against pcorr, a missing blank must have slipped in...

General tests work fine.
The doc builds fine. But the doctests fail. But it looks like the failing doctest is not related to any changes in this PR. It looks more like doctest is misinterpreting the format of the expected results.

pingouin/pairwise.py
JoElfner: Yep. In contrast: Coverage is reduced if it is not removed, since these lines won't be reached anymore (now shape[0] is always >0).

raphaelvallat:
Everything looking great then! I'm guessing the doctest fails on pairwise_ttests / pairwise_tukey functions? It's because depending on the actual screen size of your terminal, pandas will crop a certain number of columns in the middle of the dataframe. Extending your terminal to the max screen size should solve the issue and all the doctests should be running smoothly. I'm just waiting for your confirmation and I'll merge the PR right after. Thanks again!

JoElfner:
Ah yes, that sounds reasonable.
If you don't mind the remaining commented lines, feel free to merge. :)
Otherwise I'll remove the lines tomorrow.

raphaelvallat:
One last thing: could you please add an example to the notebooks/04_Correlations.ipynb file? That would be really helpful.

JoElfner:
Done! I'm ready to merge once the checks have passed. flake8 is fine, same with pytest. But the doctest still fails. May it be, that doctest sees some characters, such as the `-` in `'one-sided'` tail in Example 1 of `pairwise_corr`, as a column separator? The CI formatting, the `-` in `p-unc`, `p-corr` and `p-adjust` are imho also suspicious of breaking the doctest.

raphaelvallat:
What version of Python / Numpy / Scipy are you using? I've noticed that depending on the version there can be very small differences in float precision. I'm using Python 3.8 with Numpy 1.19.1 and Scipy 1.5.2.

JoElfner:
3.7.7, numpy 1.19.1 and scipy 1.5.2.
yeah, possibly. but distance corr is too far off.

raphaelvallat:
I'll take a look at it. Merging the PR now, thanks so much once again!

JoElfner:
Same same :) and again thanks for developing such a great tool! I'm using it quite a lot.

raphaelvallat:
Hi @JoElfner,

When you have time, would you be able to send me a screenshot of the error message for the distance_corr function (you can reach out to me at raphaelvallat9@gmail.com)? That would be helpful because I'm not able to reproduce the error.

Thanks!

PR #126, opened 2020-09-04
Implement BF formatting using the new options module

Spaak:
As discussed in #101.

The lower-level functions `bayesian.bayesfactor_ttest` etc. now return `float`s; all post-processing (i.e., formatting into scientific or positional string) is handled at the DataFrame level in `_postprocess_dataframe`.

### DIFF
--- a/pingouin/bayesian.py
+++ b/pingouin/bayesian.py
@@ -10,6 +10,8 @@ __all__ = ["bayesfactor_ttest", "bayesfactor_pearson", "bayesfactor_binom"]
 def _format_bf(bf, precision=3, trim='0'):
     """Format BF10 to floating point or scientific notation.
     """
+    if type(bf) == str:
+        return bf
     if bf >= 1e4 or bf <= 1e-4:
         out = np.format_float_scientific(bf, precision=precision, trim=trim)
     else:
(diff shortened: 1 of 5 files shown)


pingouin/bayesian.py
```diff
@@ -10,6 +10,8 @@
 def _format_bf(bf, precision=3, trim='0'):
     """Format BF10 to floating point or scientific notation.
     """
+    if type(bf) == str:
```
raphaelvallat, line 13: Why is this needed?

raphaelvallat:
Great! I just added one comment on the code. Can you also please update the changelog? Thanks!

pingouin/bayesian.py
Spaak: `pairwise.pairwise_ttests` internally calls `parametric.ttest`, which also returns a post-processed DataFrame. At the end of `pairwise_ttests`, `postprocess_dataframe` is called again, but we don't want to attempt to format the BF again (which is by then already a string).

Spaak:
> Can you also please update the changelog? Thanks!

Done!

raphaelvallat (approved):

raphaelvallat:
Justed merged the PR. Thank you!

PR #131, opened 2020-09-19
multiple regression: catch rank deficient design matrices and warn

### LINKED ISSUE #130: Rank deficiency of design matrix in multiple linear regression leads to cryptic error
Hi there, today I stumbled over an interesting error. Let me start with the example (I'll keep it close to the truth of what I actually do).

let's assume that:

- `n` are the number of trials in an experiment
- `y` are amplitudes in µV of [EEG](https://en.wikipedia.org/wiki/Electroencephalography) data at some channel/time bin (i.e., the voltage amplitude recorded at a specific EEG channel at a specific timepoint -> the same channel and timepoint per trial) 
- `X` is the design matrix --> imagine that the `n` trials in the experiment can come from 5 conditions (e.g., a number between 1 and 5 is shown as a stimulus on a screen), and each trial has a different condition ... the design matrix has as many columns as there are conditions. And in each row of the design matrix, the column corresponding to the condition of the current trial is switched to 1, the rest is 0. (and then, an intercept is added)

Ok, let's see the code:

```Python
import numpy as np
import pingouin
import statsmodels.api as sm

# make up some random data and a design matrix for multiple linear regression
n = 100 
y = np.random.randn(n)
X = np.vstack([np.random.permutation([0, 1, 0, 0, 0]) for i in range(n)])

# this will raise a ValueError
pingouin.linear_regression(X, y, add_intercept=True)

# Let's try ordinary least squares multiple regression in statsmodels instead
# This works, ... however there *is* a warning (see results.summary())
X_with_intercept = sm.add_constant(X)
model = sm.OLS(endog=y, exog=X_with_intercept)
results = model.fit()
results.summary()

# scikit learn multiple linear regression works as well
# ...
```

The ValueError is due to scipy's `lstsq` (which seems equivalent to `np.linalg.lstsq` to me), namely the rank of `X` (with added constant term) is one less than there are columns.

That in turn will lead to the residuals being unspecified, see the [docstr](https://numpy.org/doc/stable/reference/generated/numpy.linalg.lstsq.html)

> Sums of residuals; squared Euclidean 2-norm for each column in b - a*x. **If the rank of a is < N or M <= N, this is an empty array**. If b is 1-dimensional, this is a (1,) shape array. Otherwise the shape is (K,).

Emphasis by me, in our case, the rank is < N.

I have several questions:

1. Why is the design matrix rank deficient in the way I set it up?
1. Why does statsmodels OLS work, but scipy, numpy, pingouin do not work?
1. Should we perhaps improve pingouin's error message?

regarding point 3, this is the current error:

> ValueError: operands could not be broadcast together with shapes (0,) (6,) 

it'd be more helpful to say something along these lines:

> Your design matrix is rank deficient, and residuals could not be meaningfully calculated ...

sappelhoff:
closes #130 

For now I opted to only return the coefficients in case a rank deficient design matrix is detected.

The basic error of pingouin is its inability to handle the `residues` return parameter of `lstsq` whenever the design matrix is rank deficient. In that case, `residues` is an empty array `array([], dtype=float64)` instead of a float.

One option apart from warning and simply returning the coefficients only could be to "revive" the currently commented line from line 413(old) / 422(new): `    # ss_res = (resid ** 2).sum()`

Something like:

```Python
    # FIT (WEIGHTED) LEAST SQUARES REGRESSION USING SCIPY.LINALG.LSTST
    coef, ss_res, rank, _ = lstsq(Xw, yw)
    if coef_only:
        return coef
    calc_ss_res = False
    if rank < Xw.shape[1]:
        warnings.warn('Design matrix supplied with `X` parameter is rank '
                      f'deficient (rank {rank} with {Xw.shape[1]} columns. '
                      'That means that one or more of the columns in `X` '
                      'are a linear combination of one of more of the '
                      'other columns.'')
        calc_ss_res = True

    # Degrees of freedom
    df_model = rank - constant
    df_resid = n - p
    # Calculate predicted values and (weighted) residuals
    pred = Xw @ coef
    resid = yw - pred
    if calc_ss_res:
        ss_res = (resid ** 2).sum()
```

I played around with that, and it seems to be producing the same results as statsmodels. Try for yourself if you'd like:

```Python
import numpy as np 
import pingouin 
import statsmodels.api as sm 

# make up some random data and a design matrix for multiple linear regression 
n = 100  
y = np.random.randn(n) 
X = np.vstack([np.random.permutation([0, 1, 0, 0, 0]) for i in range(n)]) 

# try with pingouin (use patch from code example above)
results_pingouin = pingouin.linear_regression(X, y, add_intercept=True) 
print(results_pingouin)

# try with statsmodels
X_with_intercept = sm.add_constant(X) 
model = sm.OLS(endog=y, exog=X_with_intercept) 
results_sm = model.fit() 
print(results_sm.summary())

# compare residuals
np.allclose(results_pingouin.residuals_, results_sm.resid)  # True
np.allclose(results_pingouin.coef.to_numpy(), results_sm.params)  # True
```

Let me know what you think :-)

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -461,6 +461,7 @@ Note that this program is provided with NO WARRANTY OF ANY KIND. If you can, alw
 - `Arthur Paulino <https://github.com/arthurpaulino>`_
 - `Eelke Spaak <https://eelkespaak.nl/>`_
 - `Johannes Elfner <https://www.linkedin.com/in/johannes-elfner/>`_
+- `Stefan Appelhoff <https://stefanappelhoff.com>`_
 
 How to cite Pingouin?
 =====================
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -9,6 +9,7 @@ v0.3.9 (dev)
 **Enhancements**
 
 a. :py:func:`pingouin.plot_paired` now supports an arbitrary number of within-levels as well as horizontal plotting. See `PR 133 <https://github.com/raphaelvallat/pingouin/pull/133>`_.
+b. :py:func:`pingouin.linear_regression` now handles a rank deficient design matrix X by producing a warning and trying to calculate the sum of squared residuals without relying on :py:func:`np.linalg.lstsq`. See `issue 130 <https://github.com/raphaelvallat/pingouin/issues/130>`_.
 
 v0.3.8 (September 2020)
 -----------------------
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -512,6 +512,7 @@ Note that this program is provided with NO WARRANTY OF ANY KIND. If you can, alw
 - `Arthur Paulino <https://github.com/arthurpaulino>`_
 - `Eelke Spaak <https://eelkespaak.nl/>`_
 - `Johannes Elfner <https://www.linkedin.com/in/johannes-elfner/>`_
+- `Stefan Appelhoff <https://stefanappelhoff.com>`_
 
 How to cite Pingouin?
 =====================
--- a/pingouin/regression.py
+++ b/pingouin/regression.py
@@ -406,7 +406,7 @@ def linear_regression(X, y, add_intercept=True, weights=None, coef_only=False,
     calc_ss_res = False
     if rank < Xw.shape[1]:
         warnings.warn('Design matrix supplied with `X` parameter is rank '
-                      f'deficient (rank {rank} with {Xw.shape[1]} columns. '
+                      f'deficient (rank {rank} with {Xw.shape[1]} columns). '
                       'That means that one or more of the columns in `X` '
                       'are a linear combination of one of more of the '
                       'other columns.')
--- a/pingouin/regression.py
+++ b/pingouin/regression.py
@@ -399,12 +399,14 @@ def linear_regression(X, y, add_intercept=True, weights=None, coef_only=False,
         Xw = X
         yw = y
 
-    # FIT (WEIGHTED) LEAST SQUARES REGRESSION USING SCIPY.LINALG.LSTST
+    # FIT (WEIGHTED) LEAST SQUARES REGRESSION USING NP.LINALG.LSTST
     coef, ss_res, rank, _ = np.linalg.lstsq(Xw, yw, rcond=None)
+    ss_res = ss_res[0] if ss_res.shape == (1,) else ss_res
     if coef_only:
         return coef
     calc_ss_res = False
     if rank < Xw.shape[1]:
+        # in this case, ss_res is of shape (0,), i.e., an empty array
         warnings.warn('Design matrix supplied with `X` parameter is rank '
                       f'deficient (rank {rank} with {Xw.shape[1]} columns). '
                       'That means that one or more of the columns in `X` '
@@ -415,6 +417,7 @@ def linear_regression(X, y, add_intercept=True, weights=None, coef_only=False,
     # Degrees of freedom
     df_model = rank - constant
     df_resid = n - rank
+
     # Calculate predicted values and (weighted) residuals
     pred = Xw @ coef
     resid = yw - pred
--- a/pingouin/regression.py
+++ b/pingouin/regression.py
@@ -400,7 +400,7 @@ def linear_regression(X, y, add_intercept=True, weights=None, coef_only=False,
         yw = y
 
     # FIT (WEIGHTED) LEAST SQUARES REGRESSION USING SCIPY.LINALG.LSTST
-    coef, ss_res, rank, _ = np.linalg.lstsq(Xw, yw)
+    coef, ss_res, rank, _ = np.linalg.lstsq(Xw, yw, rcond=None)
     if coef_only:
         return coef
     calc_ss_res = False
--- a/pingouin/regression.py
+++ b/pingouin/regression.py
@@ -4,7 +4,7 @@ import numpy as np
 import pandas as pd
 import pandas_flavor as pf
 from scipy.stats import t, norm
-from scipy.linalg import pinv, pinvh, lstsq
+from scipy.linalg import pinv, pinvh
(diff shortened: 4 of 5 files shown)


raphaelvallat:
Thanks for the PR @sappelhoff! I think I like the second approach better because it would preserve the final output (a dataframe). Could you please:

- Push to the develop branch (working version) instead of the master (for release only)
- Add unit testing against statsmodels when X is rank deficient in tests/test_regression.py
- Add your change to the docs/changelog.rst
- If you have any insights from code / paper on what is the most adequate way to calculate the residuals when design matrix is rank-deficient, that'd be fantastic. It's good to know that calculating the residuals as if the design matrix was full rank gives similar to statsmodels, but I'd love to understand why it is the case!

Let me know if you have any questions/issues with any of these steps,

Thanks again!

sappelhoff:
Will do, next week :+1:

sappelhoff:
> Push to the develop branch (working version) instead of the master (for release only)

done

> Add unit testing against statsmodels when X is rank deficient in tests/test_regression.py

done

> Add your change to the docs/changelog.rst

done, and I took the liberty to add my name to the list of contributors. Hope that was fine.

> If you have any insights from code / paper on what is the most adequate way to calculate the residuals when design matrix is rank-deficient, that'd be fantastic.

no idea, unfortunately

---

Some results of the tests:

- It seems the adjusted R squared value is very different between Statsmodels and Pingouin. 
- The coefficients, residuals, and R squared (not adjusted) are very similar.
- All other parameters are "reasonably" similar, but with slight deviations of up to 0.05 difference.

Do you know why adusted R squared is so different?

--- 

one more thing:

@raphaelvallat I noticed that in pingouin there are some calls of `np.linalg.lstsq`, and in regression.py, there is one for `scipy.linalg.lstsq` ... to my knowledge (and according to a quick check) both function produce the exact same results --> I think I would replace the `scipy.linalg.lstsq` call with the numpy one (i) for consistency reasons, and (ii) because numpy is more standard than scipy

raphaelvallat:
> done, and I took the liberty to add my name to the list of contributors. Hope that was fine.

Absolutely, that was actually going to be my next action item :-)

> It seems the adjusted R squared value is very different between Statsmodels and Pingouin.

This only when the X matrix is rank-deficient, correct?

> All other parameters are "reasonably" similar, but with slight deviations of up to 0.05 difference.

Hmm, what parameters? I can dive into the code this weekend, we really need to reach a perfect agreement with statsmodel.

Finally, there is a conflict in the changelog because I merged the other PR, could you please make sure you're using the latest version of the develop branch and resolve the conflict?

Thanks again!

raphaelvallat:
Oh and please feel free to use NumPy instead of SciPy for LSTSQ!

sappelhoff:
> Finally, there is a conflict in the changelog because I merged the other PR, could you please make sure you're using the latest version of the develop branch and resolve the conflict?

done

> Hmm, what parameters? I can dive into the code this weekend, we really need to reach a perfect agreement with statsmodel.

I documented that in the tests ... and I agree that perfect agreement must be the requirement ... and if that cannot be reached we need to know why not.

> This only when the X matrix is rank-deficient, correct?

currently the tests are only against statsmodels for rank-deficient X ... perhaps we can/should extend them to also cover full-rank X

raphaelvallat:
@sappelhoff sorry I didn't get the chance to dive into this issue this weekend, but something that came to my mind that may explain the discrepancies is that perhaps we should update the `p` variable (from `n, p = X.shape[0], X.shape[1]`) to be equal to rank instead. So let's say we have 5 columns in our dataframe, but the rank is only 4, then `p` should be equal to 4. This will impact the `df_resid` variable which in turn impacts the adj R^2, p-value, 95%CIs, etc. 

TLDR: what happens if you replace `df_resid = n - p` by `df_resid = n - rank`?

Thanks!

sappelhoff:
> what happens if you replace df_resid = n - p by df_resid = n - rank?

that's a good hunch you had -> it fixes the discrepancies (at least locally, let's wait for the CI) :slightly_smiling_face: 

so now when pingouin encounters a rank deficient matrix it does NOT error out anymore (as previously), but instead emits a warning and continuous to produce results equivalent to what statsmodels would produce.

the factual code changes were:

1. calculating sum of squared residuals "manually" when `lstsq` returns nothing due to rank-deficiency
1. calculating residual degrees of freedom with n and *rank* instead of n and *n_columns*

sappelhoff:
@raphaelvallat there are some tests failing, but they seem to be unrelated to my changes.

On a related note: What would you think of switching the CI servive from Appveyor to GitHub Actions?

I have had good experiences with GH Actions on several repos: Very easy to set up, testing on Win,Mac,Linux for free, runs quickly, runs reliably --> and of course ... the GitHub user interface integration is good :)

I am saying this because Travis does not show up on this PR although it's configured (that's a recent problem I have observed on several repos) ... and because Appveyor is usually the slowest of all CI services I encounter (and the markup of the Appveyor output is not well readable IMHO)

raphaelvallat:
Hi @sappelhoff,

Hm not sure to understand why the CI tests failed but they all seem to be related to linear_regression function. Have you tried running the tests on the docstring examples with `pytest --doctest-modules`?

Oh yes please, I'd love to get rid of AppVeyor! It makes me genuinely anxious every time I push a new commit... I'm ok with Travis, but if there's a better way to centralize all the unit tests with GitHub Actions, then I'm all in!

sappelhoff:
mh yeah, you are right. Given that #137 passes the tests, there must be something in this PR that causes the failures. But it's a bit puzzling.

raphaelvallat:
Great! All tests are passing so I'm merging the issue now. Thanks so much again for your contribution!

sappelhoff:
yay :-) thanks for your help!

PR #133, opened 2020-09-24
plot_paired for n>2 subjects and with horizontal orientation

### LINKED ISSUE #132: plot_paired for n>2 subjects
Hi @raphaelvallat 

are you interested in an implementation of a paired plot for an arbitrary number of unique subjects?

This could look like:
evolution of income over subjects

JoElfner:
Closes #132 

This adds the functionality to make paired plots with an arbitrary number of subjects/orders. I refactored some code to not need any loops when `len(x_cat) == 2`, else looping over subjects is required. For this, `sns.pointplot` now also takes the `hue` and `palette` (replacing color) arguments. To disable the default legend displayed with hue, it is removed afterwards. Furthermore the z-order of the pointplot is set to two, since otherwise each `hue` has an increasing z-order.

Additionally I added the `orient` argument, which allows for rotating the plot horizontally. This is imho favorable for `len(x_cat) > 3`.

Tests, doc and example will follow. Any remarks?


raphaelvallat:
Wow that was fast!! The code looks very clean. In addition with the test, doc and example, could you just submit to the develop branch please?

Thanks again!

JoElfner:
Yeah, already had it implemented locally, so only had to adjust it for pingouin. :)
But I've got a problem now: The implementation works for n==2 and n>3 unique within-levels. The bug seems to be within seaborn pointplot, when the `order` argument has a length different from the number of of categoric levels, here len(x_cat):
- When `order` passed to pointplot is the current combination of withins, f.i. `order=['August', 'June']`, then seaborn assumes there are only two levels in total, setting the range of the categoric variable to 0-1. This is also done in the next iteration, when order is f.i. `order=['June', 'January']`, with the result that the lines/points of this combination are also plotted in the 0-1 range, thus all pointplot elements are at the same within-level.
- When `order` passed to pointplot contains the order of all levels, f.i. `order=['August', 'June', 'January']` and `len(order) == 3`, seaborn color handling is buggy. The colors it passes to matplotlib scatter plot is `facecolor=(r, g, b), edgecolor=(r, g, b)`, resulting matplotlib to interpret each float r/g/b as a single color value, since `len((r, g, b)) == len(order)`. Changing the seaborn color handling to: `facecolor=((r, g, b), ), edgecolor=((r, g, b), )` (making all colors explicitly to tuples of rgb-values) could probably solve it. But this would require waiting for seaborn to release a new version and keep n==3 levels deprecated in pingouin until that point.

I could give it a try by either replacing pointplot with lineplot and moving some of the categoric handling done in pointplot to plot_paired OR by using matplotlib directly. What do you think?

JoElfner:
Well... This got much more complicated than I thought...
I switched to lineplot and scatterplot, allowing for a more flexible integration of different levels of the `within`-factor. This took some tweaks to the plot style to make it look exactly like that of pointplot.
Furthermore ordering of numeric `within` columns had to be done by a replacement-within-column.
I tested it extensively with all possible combinations of values and it is working fine. You just have to decide, if you want to adapt it the way it is.

raphaelvallat:
Thanks for looking into this. I'll take a deep dive tomorrow. For reference, would you be able to upload the figures generated by the examples in the docstring here?

JoElfner:
Sure

Default paired plot:
```
import pingouin as pg
df = pg.read_dataset('mixed_anova').query("Time != 'January'")
df = df.query("Group == 'Meditation' and Subject > 40")
ax = pg.plot_paired(data=df, dv='Scores', within='Time',
                    subject='Subject', dpi=150)
```
doctest_default


Paired plot on an existing axis (no boxplot and uniform color):
```
df = pg.read_dataset('mixed_anova').query("Time != 'January'")
df = df.query("Group == 'Meditation' and Subject > 40")
fig, ax1 = plt.subplots(1, 1, figsize=(5, 4))
pg.plot_paired(data=df, dv='Scores', within='Time',
                subject='Subject', ax=ax1, boxplot=False,
                colors=['grey', 'grey', 'grey'])  # doctest: +SKIP
```
doctest_no_boxplot


Horizontal paired plot with three unique within-levels:
```
df = pg.read_dataset('mixed_anova').query("Group == 'Meditation'")
# df = df.query("Group == 'Meditation' and Subject > 40")
pg.plot_paired(data=df, dv='Scores', within='Time',
               subject='Subject', orient='h')  # doctest: +SKIP
```
doctest_3levels_horizontal


With the boxplot on the foreground:
```
df = pg.read_dataset('mixed_anova').query("Time != 'January'")
df = df.query("Group == 'Control'")
ax = pg.plot_paired(data=df, dv='Scores', within='Time',
                    subject='Subject', boxplot_in_front=True)
```
doctest_boxplot_in_front

raphaelvallat (approved):

raphaelvallat:
Approving the PR now, thanks again for your contributions @JoElfner !

JoElfner:
Thanks for merging :)

PR #137, opened 2020-10-07
INFRA: add GitHub Actions workflow for pytest (replace Appveyor)

sappelhoff:
see: https://github.com/raphaelvallat/pingouin/pull/131#issuecomment-704555993

I did not remove Appveyor yet to let you see the new setup and approve of it @raphaelvallat -> you could then remove appveyor or GitHub Actions again, after deciding for one over the other.

Note that I kept the CI script very simple, because Appveyor also did only "pytest", and not formatting checks, etc.

I am not sure whether or not this will directly work, or whether it'll have to be merged first :man_shrugging: 

edit: will have to be merged first :slightly_smiling_face:


raphaelvallat:
This is really amazing!! Very clean config file too. One question I have is can we link this to codecov? Similar to what I'm doing at the end of the Travis config file:

```yaml
after_success:
  - codecov
```

Also, if possible would love to add the doctest-modules here, so something like: 

```yaml
- name: Test with pytest
      run: |
       pytest --doctest-modules --verbose
       codecov
```

sappelhoff:
is this what you had in mind @raphaelvallat ?

raphaelvallat:
That looks great. I'll try it later today and let you know if I run into any issues!

PR #144, opened 2020-11-13
Add F test method to friedman function

PuncocharM:
Possible issue: I'm not sure how to test the new method. There are implementations of Kendall's W in other software (e.g. R), but I didn't find the F test implemented, this might be the first one.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -15,6 +15,7 @@ See `issue 139 <https://github.com/raphaelvallat/pingouin/issues/139>`_.
 
 a. :py:func:`pingouin.plot_paired` now supports an arbitrary number of within-levels as well as horizontal plotting. See `PR 133 <https://github.com/raphaelvallat/pingouin/pull/133>`_.
 b. :py:func:`pingouin.linear_regression` now handles a rank deficient design matrix X by producing a warning and trying to calculate the sum of squared residuals without relying on :py:func:`np.linalg.lstsq`. See `issue 130 <https://github.com/raphaelvallat/pingouin/issues/130>`_.
+c. :py:func:`pingouin.friedman` now has an option to choose between Chi square test or F test method.
 
 v0.3.8 (September 2020)
 -----------------------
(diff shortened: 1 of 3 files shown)


raphaelvallat:
Hi @PuncocharM !

Thanks for the PR! The code is looking great, just a few comments on documentation/testing:

1) Got it for unit testing of the F-test. I'm not too worried as long as we follow the exact same steps described in the paper. However, could you please add a unit test for the W-statistic? 

2) Could you please add an example with the F-test in the docstring of the function ("Example" section)?

4) _"Instead they recommend the F test, which has the correct size and behaves like a permutation test."_ Is that from the original paper? I think this sentence is slightly confusing, especially the first part. Do you mean that the test behaves well even with small sample sizes?

5) Could you please add your modifications to the docs/changelog.rst file ("What's new")?

6) Please make sure to use space between mathematical operators (e.g. `W * (n - 1) / (1 - W)` instead of `W * (n-1) / (1-W)`) so that the code follows the PEP8 guidelines.

Thanks again!
Raphael

PuncocharM:
Hi @raphaelvallat !

Sorry for the delay, I addressed your comments with the commit above.

>     1. Got it for unit testing of the F-test. I'm not too worried as long as we follow the exact same steps described in the paper. However, could you please add a unit test for the W-statistic?
I agree that this can and should be tested. I added the test comparing it to a result from R.

>     2. Could you please add an example with the F-test in the docstring of the function ("Example" section)?
Done.

>     3. _"Instead they recommend the F test, which has the correct size and behaves like a permutation test."_ Is that from the original paper? I think this sentence is slightly confusing, especially the first part. Do you mean that the test behaves well even with small sample sizes?

Citing the paper:

> It is shown that the Friedman test is too conservative and less powerful than both the F test and the permutation test for concordance which always have a correct size and behave alike. The F test should be preferred because it is computationally much easier.

The point is that both Chi squared and F are asymptotical approximations, but the Chi squared is too conservative compared to permutation test and F test, which are both correct size. F test is superior to permutation test in computational complexity, there fore is the preferred one from all the three. I can rephrase the sentence in the documentation if you tell me how.


>     4. Could you please add your modifications to the docs/changelog.rst file ("What's new")?
Done. I'm not sure if I should go into more details there?

>     5. Please make sure to use space between mathematical operators (e.g. `W * (n - 1) / (1 - W)` instead of `W * (n-1) / (1-W)`) so that the code follows the PEP8 guidelines.
I fixed it. Although now it's according to PEP8, I don't think it's more readable. Anyway it's a small issue.

Let me know if there is anything else you'd like me to modify or add.

raphaelvallat (approved):

raphaelvallat:
Fantastic, thank you @PuncocharM for the PR, merging now!

PR #150, opened 2021-01-14
DOC: clarify kendall's tau "B", add GH build artifacts, add sphinx-copybutton, color-output

sappelhoff:
I was using the corr function and noticed that it's unclear which kendall's tau is being used.

However, when starting the PR I saw deeper down in the documentation string, that you actually do say that.

I still suggest to add the information already in the parameter documentation.

PS: It seems like there is no CI step that checks whether the docs are cleanly built. Is that correct? I added such a build step. It will run only for ubuntu-latest and python 3.8, so it will not draw too many resources or cost a lot of time. At the end of the build, the docs can be downloaded as a zip file and inspected. that can be helpful for contributors who don't have the full docs requirements installed but still want to check the impact of their changes on the docs.


sappelhoff:
the artifacts can be downloaded after clicking on the gh actions "check", see screenshot:

image

(I am not trying to be annoying here, just assuming that you don't know this already :laughing: sorry if you do.)

I can then see the result of my change:

image

raphaelvallat:
Hi @sappelhoff ! Thank you so much, this is fantastic, especially the doc checking (and I didn't know..!). Just a question before I merge, what's the -C for in `make -C docs html`? Also, could you perhaps add the screenshot to the contributing.rst page of the documentation? That way, future collaborators will know that they can check the doc without having to compile it on their own computers? Thanks again!

sappelhoff:
> what's the -C for in make -C docs html

I added a brief section about that to CONTRIBUTING as well --> it basically tells make to first **C**hance to the directory after `-C`, then execute the recipe, and *then return to the original directory*.

> Also, could you perhaps add the screenshot to the contributing.rst page of the documentation? That way, future collaborators will know that they can check the doc without having to compile it on their own computers?

Done!

sappelhoff:
I added two more changes that you hopefully agree are useful:

1. adding the[ "sphinx-copybutton"](https://github.com/executablebooks/sphinx-copybutton), which allows users to simply "copy" each code cell using a tiny icon (see screenshot)

image

2. forcing the GitHub actions log to be colorful

raphaelvallat:
Wonderfull!! Thanks again! I'll merge now and then add the PR to the changelog.

sappelhoff:
Glad you liked it, thanks for your continued support of the package!

PR #156, opened 2021-02-12
Internal qsturng Implementation Substitution

### LINKED ISSUE #155: qsturng.py Inaccuracy
Hello,

I was doing some testing recently, and noticed that the `qstrung.py` implementation in Pingouin seems to give a large error when calling pstrung. The implementation, which seems almost identical to Statsmodels', gives errors on the order of 10^-2 when compared with manual integral evaluation, R, and Statmodels.

Here's a quick test script to demonstrate what I mean: 
```python
from pingouin.external.qsturng import psturng as ping_q
from statsmodels.stats.libqsturng import psturng as stats_q

import numpy as np
import itertools

q_i = np.arange(1, 4)
k_i = range(2, 6)
v_i = range(2, 120, 10)

tot_err = 0
total = 0

print("Calculating average error")
for q, k, v in itertools.product(q_i, k_i, v_i):
	tot_err += abs(stats_q(q, k, v) - ping_q(q, k, v))
	total += 1

print(f"Average stock error: {tot_err / total}")
```
```
> Calculating average error
> Average stock error: [0.015114]
```

It looks like the problem is that somewhere along the way, some `min()` statements got switched out for some `max()` ones.

Pingouin:
```python
    return math.sqrt(2) * -y * \
           scipy.stats.t.isf((1. + p) / 2., max(v, 1e38))
```

Statsmodels:
```python
    v = min(v, 1e38)
    return math.sqrt(2) * -y * scipy.stats.t.isf((1. + p) / 2., v)
```

The mismatches are on **lines 580, 582, 592, 725**
Simply replacing all the `max()` statements with `min()` statements seems to fix the problem. 
```
> Calculating patched error
> Average patched error: [0.]
```

Let me know if you'd like me to submit a PR. 

Thanks!

DominicChm:
This PR resolves #155 by removing the internal `qsturng` and `psturng` function implementations (used for tukey's pairwise comparison) and replacing them with statsmodels' version.

The below is a worst-case example of the behavior seen with the current version. Pingouin's error compared to Statsmodels' is much greater than it should be, considering that the implementation of their `psturng` and `qsturng` functions is almost identical. With this change, that implementation is replaced with Statmodels' more accurate one.

cdf-0 01-8 225917112372644-20-10_DATA
Note: This graph is of the error of various different functions at varying degrees of input precision. In this case, statmodels' and pingouin's lines are generated by calling their `psturng` functions with the labeled input parameters, then subtracted from the high-precision numerical result of the integral the `psturng` function emulates to get error.


raphaelvallat:
Thanks @DominicChm, that's really great! A few minor comments before merging:

1) Can you add the changes to the docs/changelog.rst?
2) Can you also update the Examples section in the documentation of the pairwise_tukey and pairwise_gameshowell function?
3) Finally, make sure that all the flake8 and doctest run smoothly by following the instructions here: https://pingouin-stats.org/contributing.html

DominicChm:
Ok, @raphaelvallat, I've updated examples and changelog in a way that I hope is consistent with previous ones. I added the change to the 0.3.9 (current?) version because I'm not sure about when this change should ship, but could start another section instead if it makes more sense.

raphaelvallat:
Amazing @DominicChm ! So, we actually need to start a new section, I think let's go for 0.3.10 as I'm hoping to release this bugfix soon. Thanks again!

DominicChm:
@raphaelvallat No problem, happy to help! I've added the new changelog section. I included the same disclaimer from the previous bugfix because it seems like this problem could significantly affect p-values.

raphaelvallat (approved):

raphaelvallat:
Merging now, thanks again!

PR #160, opened 2021-02-25
[BUGFIX] Add safety check for `tail` argument in corr func

sappelhoff:
I noticed a missing safety check in the corr function and add it in this PR.

#### does not raise an issue (notice the wrong `tail` arg)

```python
stat_bnt = pingouin.correlation.corr(
    _curdat[colname], _curdat["correct_choice"], method="kendall", tail="two-tailed"
)
```

#### raises an issue (`tail` arg still wrong, but corr method changed)

```python
stat_bnt = pingouin.correlation.corr(
    _curdat[colname], _curdat["correct_choice"], method="pearson", tail="two-tailed"
)
```

So the wrong input was sometimes still called, but not always. With my changes, it is always called.

This probably fixes a silent bug: When users used `kendall` as the `method`, and specified `two-tailed` (instead of the correct `two-sided`), no error was raised ... and a one-sided p value was returned :skull: :grimacing: 

see:

https://github.com/raphaelvallat/pingouin/blob/0c3bf73629ffb1826dc298f0ee498e0784222b5e/pingouin/correlation.py#L556

if it's "two-tailed", it's treated the same way as "one-sided", and the missing safety check (now added in this PR) prevented the wrong "two-tailed" argument from being caught early.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -3,6 +3,15 @@
 What's new
 ##########
 
+v0.3.11 (unreleased)
+--------------------
+
+**Bugfixes**
+
+a. Passing a wrong ``tail`` argument to :py:func:`pingouin.correlation.corr` now *always* raises an error (see `PR 160 <https://github.com/raphaelvallat/pingouin/pull/160>`_).
+   In previous versions of ``pingouin``, using any ``method`` other than ``"pearson"`` and a wrong ``tail`` argument such as ``"two-tailed"`` or ``"both"``
+   (instead of the correct ``"two-sided"``) may have resulted in silently returning a one-sided p-value.
+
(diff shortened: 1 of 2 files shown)


raphaelvallat:
Thanks for your sharp eyes @sappelhoff! Would you be able to add this to the _docs/changelog.rst_ file? You will need to start a new section (version 0.3.11). Make sure you add a hyperlink to the PR. Thanks again!

sappelhoff:
sure! Let me know if more is needed.

raphaelvallat (approved):

raphaelvallat:
Merging now! Thanks again (and love the branch name btw 😁 )

PR #175, opened 2021-05-27
boxM implementation

GanshengT:
A generic implementation of Box's M test. Concerning maths expressions would be added later.


raphaelvallat:
Hi @GanshengT,

Thanks for the PR! I just did a very short scan, and I have three main comments for now:

1. We need to add test against R or Matlab in [pingouin/tests/test_multivariate.py](https://github.com/raphaelvallat/pingouin/blob/master/pingouin/tests/test_multivariate.py). More broadly, please make sure to follow the guidelines for contributing: https://pingouin-stats.org/contributing.html
2. Does the Box test only work for covariance matrices of similar shapes? By using a 3D numpy array as input, we're forcing all the covariance matrices to have similar shapes. If desired, we could have a more arbitrary input, as I've done in [pingouin.homoscedasticity](https://pingouin-stats.org/generated/pingouin.homoscedasticity.html#pingouin.homoscedasticity)
3. There shouldn't be any uppercase in function names in Python, so I would suggest renaming the function to `box_test`, or `box_m`

Raphael

GanshengT:
Thanks for your comments @raphaelvallat.  
1. I am on top of that.
2. To the best of my knowledge, the Box's M test is used to test covariance matrices of the same shape because the dependent/independent variables should be the same across groups. If the Box's M test can be generalized to test any covariance matrices, please let me know. In this case, how do we define the equality of covariance matrices of different shapes?
3. One of my bad habits. `box_m` is a good idea.

pingouin/multivariate.py
```diff
@@ -326,22 +326,20 @@ def boxM (covs,sizes, alpha=.001):
...
-        
+            T=-T+(1/(sizes[idx_cov]-1))
+        c=-k1*(T-(1/(np.sum(sizes)-covs.shape[0])))      
     # calculate U statistics and degree of fredom
     u =-2*(1-c)*np.log(M)
+    
     df = 0.5*covs.shape[1]*(covs.shape[1]+1)*(covs.shape[0]-1)
     p = 1-scipy.stats.chi2.cdf(u,df)
     equal_cov = True if p > alpha else False
     stats = pd.DataFrame(data={'Chi2':[u], 'df':[df],'pval':[p],'equal_cov':[equal_cov]},index=["Box's M"])
```
raphaelvallat, line 343: Could you please rename the index to `box` instead of `Box's M`? We want to avoid having "'" in index names.

raphaelvallat:
You are right for 2), covariance matrices should be the same size.

Now that we have this working implementation, I think it would be great to change the input such that users specify the dependent variables and grouping variable instead (`dvs` and `group`, similar to https://pingouin-stats.org/_modules/pingouin/distribution.html#homoscedasticity). I think using pandas.DataFrame.groupby.cov() we could greatly simplify the user's input (i.e. no need to pass each unique covariance matrix and/or the sample size).

Let me know what you think!

pingouin/multivariate.py
```diff
@@ -1,5 +1,6 @@
 import numpy as np
 import pandas as pd
+import scipy
```
raphaelvallat, line 3: Please avoid generic call to scipy and instead import the specific function (from scipy.stats import chi2)

pingouin/multivariate.py
```diff
@@ -254,3 +255,91 @@ def multivariate_ttest(X, Y=None, paired=False):
...
+        sizes = np.array(sizes)
+        
+    # calculate pooled S and M statistics
+    ## covs.shape[0] is the number of covariance matrices
+    ## covs.shape[1] is the number of variables
+    ## np.sum(sizes) is the total number of observations
+    E = np.zeros(covs[0].shape)
+    M=1
+    for idx_cov in range(covs.shape[0]):
+        E+=(sizes[idx_cov]-1)*covs[idx_cov]
```
raphaelvallat, line 321: Please make sure to follow flake8 code guidelines (e.g. there should always be space between numeric operators, a + b = 2 instead of a+b=c)

GanshengT:
> You are right for 2), covariance matrices should be the same size.
> 
> Now that we have this working implementation, I think it would be great to change the input such that users specify the dependent variables and grouping variable instead (`dvs` and `group`, similar to https://pingouin-stats.org/_modules/pingouin/distribution.html#homoscedasticity). I think using pandas.DataFrame.groupby.cov() we could greatly simplify the user's input (i.e. no need to pass each unique covariance matrix and/or the sample size).
> 
> Let me know what you think!

Good idea!

GanshengT:
I implemented a new version of ```box_m``` that takes Dataframe, variables, and group as inputs. It currently works only for long format.

pingouin/multivariate.py
```diff
@@ -254,3 +254,102 @@ def multivariate_ttest(X, Y=None, paired=False):
     stats = {'T2': t2, 'F': fval, 'df1': df1, 'df2': df2, 'pval': pval}
     stats = pd.DataFrame(stats, index=['hotelling'])
     return _postprocess_dataframe(stats)
+
+def box_m (data, vars=None, group=None, alpha=.001):
```
raphaelvallat, line 258: @GanshengT  For consistenty with Pingouin other functions, can you please rename `vars` to `dvs` (= dependent variables)

raphaelvallat, line 258: Is there a reason why the significance level is set to 0.001? Is that standard for the Box M test?

pingouin/multivariate.py
```diff
@@ -254,3 +254,102 @@ def multivariate_ttest(X, Y=None, paired=False):
...
+    from scipy.stats import chi2
+    """Test equality of covariance matrices.
+
+    Parameters
+    ----------
+    data : :py:class:`pandas.DataFrame`, list or dict
+        Iterable. Can be either a list / dictionnary of iterables
```
raphaelvallat, line 265: This is not really clear. Data can be either a long-format Pandas dataFrame, or an iterable (e.g. list, dict), correct?

pingouin/multivariate.py
```diff
@@ -254,3 +254,102 @@ def multivariate_ttest(X, Y=None, paired=False):
...
+        assert vars is not None and group is not None, 'data should be in the long-format'
+        # Long-format Pandas DataFrame
+        assert group in data.columns
+        assert set(vars).issubset(data.columns)
+        grp = data.groupby(group, observed=True)[vars]
+        assert grp.ngroups > 1, 'Data must have at least two columns.'
+        covs = grp.cov()
+        num_covs,num_vars = covs.index.levshape
+        sizes = grp.count().iloc[:,0]        
+
```
raphaelvallat, line 324: Please avoid double blank lines. Can you please check that the code follows the flake8 guidelines? Thanks!

pingouin/multivariate.py
```diff
@@ -254,3 +254,102 @@ def multivariate_ttest(X, Y=None, paired=False):
...
+        
+    # calculate pooled S and M statistics
+    ## covs.shape[0] num_covs is the number of covariance matrices
+    ## covs.shape[1] num_vars is the number of variables
+    ## np.sum(sizes) is the total number of observations
+        E = np.zeros([num_vars,num_vars])
```
raphaelvallat, line 330: Why the indent here?

pingouin/tests/test_multivariate.py
```diff
@@ -66,3 +68,30 @@ def test_multivariate_ttest(self):
         stats = multivariate_ttest(X_na, Y, paired=True)
         assert stats.at['hotelling', 'df1'] == 3
         assert stats.at['hotelling', 'df2'] == X.shape[0] - 1 - X.shape[1]
+
+    def test_box_m(self):
+        """Test function box_m.
+        Tested against the R package biotools (iris dataset).
+        """
+        
+        iris = datasets.load_iris()
```
raphaelvallat, line 77: Please remove duplicated lines

pingouin/tests/test_multivariate.py
```diff
@@ -66,3 +68,30 @@ def test_multivariate_ttest(self):
...
+        
+        # test for generic box_m
+        ##calculate covariance matrices
+        #cov_target0 = df[df['target']==0].iloc[:,:4].cov().values
+        #cov_target1 = df[df['target']==1].iloc[:,:4].cov().values
+        #cov_target2 = df[df['target']==2].iloc[:,:4].cov().values
+        #calculate the sample size for each 'group'
+        #sizes = [df[df['target']==0].shape[0],df[df['target']==1].shape[0],df[df['target']==2].shape[0]]
+        #stats = box_m(np.array([cov_target0,cov_target1,cov_target2]),sizes)
+        assert round(stats.at["Box's M", 'Chi2'], 5) == 140.94305
```
raphaelvallat, line 94: Why is the Chi2 value rounded to 5? Does the output of R have more float precision?

raphaelvallat (commented):
Hi @GanshengT,

Fantastic, thanks! I've added a few minor comments to the code.

pingouin/multivariate.py
GanshengT: 0.001 is commonly used. I cite:
> In addition, Box’s M has very little power (Cohen, 2008) for small sample sizes; if your small-sample result is not significant, it doesn’t necessarily indicate that the covariance matrices are equal. The test has also been criticized for being overly sensitive for large sample sizes. In other words, it will report a statistically significant result when one doesn’t actually exist. To address this particular issue, a smaller alpha level (e.g. .001) is recommended (Hahs-Vaughn, 2016).

The reference is **Hahs-Vaughn, D. (2016). Applied Multivariate Statistical Concepts. Taylor & Francis.**

pingouin/tests/test_multivariate.py
GanshengT: The precision of the output of box's M depends on the package of the function (i.e. MVTests, biotools, etc.) For example, in biotools, the Chi2 value is rounded to 2.

pingouin/multivariate.py
GanshengT: Could you explain more in detail how to run flake8 checks? The instruction on the contribution page is not so clear.

pingouin/multivariate.py
```diff
@@ -254,3 +254,97 @@ def multivariate_ttest(X, Y=None, paired=False):
     stats = {'T2': t2, 'F': fval, 'df1': df1, 'df2': df2, 'pval': pval}
     stats = pd.DataFrame(stats, index=['hotelling'])
     return _postprocess_dataframe(stats)
+
+def box_m (data, dvs=None, group=None, alpha=.001):
+    from scipy.stats import chi2
```
raphaelvallat, line 259: @GanshengT Can you please change the import to after the docstring?

pingouin/multivariate.py
raphaelvallat: You can install the flake8 package and then analyze your code with `flake8` command. See more details here: https://pypi.org/project/flake8/

raphaelvallat:
Hi @GanshengT,

Sorry for the delay. The PR looks good. A few more comments:

1) Before I merge, could you add a second test against another dataset (e.g. the "penguins" dataset in Pingouin)? 

2) Does the current function handle unequal sample size in the grouping variable and/or missing values? Would be good to test these cases against R.

3) Feel free to add your change to the docs/changelog.rst file, with link to the current PR.

Thanks,
Raphael

raphaelvallat:
Also it seems that the CI tests are failing, can you please fix this? Thank you!

GanshengT:
> Hi @GanshengT,
> 
> Sorry for the delay. The PR looks good. A few more comments:
> 
> 1. Before I merge, could you add a second test against another dataset (e.g. the "penguins" dataset in Pingouin)?
> 2. Does the current function handle unequal sample size in the grouping variable and/or missing values? Would be good to test these cases against R.
> 3. Feel free to add your change to the docs/changelog.rst file, with link to the current PR.
> 
> Thanks,
> Raphael

1. Of course. 
2. The current function handles unequal sample size, but it does not handle the missing value. Any thoughts?
3. Sure.

GanshengT:
The tests should work fine. However, the dataset penguins can not be tested against R as the function in library "biotools" does not handle NAs neither.

raphaelvallat:
Hi @GanshengT,

Thanks for all that. I think it's okay that the function cannot handle NaN if it's also the same behavior in R, we should just add a note in the documentation. I think there are some conflicts caused by the latest version of Pingouin that I released a few weeks ago. Could you please fix the conflicts before I merge?

Thanks,
Raphael

GanshengT:
The conflicts have been solved

pingouin/multivariate.py
```diff
@@ -254,3 +254,114 @@ def multivariate_ttest(X, Y=None, paired=False):
...
+    .. [1] Rencher, A. C. (2003). Methods of multivariate analysis (Vol. 492).
+    John Wiley & Sons.
+
+    Examples
+    --------
+    1. Box's M testing 3 covariance matrices from 'tip' dataset
+
+    >>> import numpy as np
+    >>> import pingouin as pg
+    >>> import scipy.stats
```
raphaelvallat, line 310: Please removed unused call to scipy.stats

pingouin/multivariate.py
```diff
@@ -254,3 +254,114 @@ def multivariate_ttest(X, Y=None, paired=False):
...
+    >>> data = pg.read_dataset('tips')[['total_bill','tip','size']]
+    >>> box_m(data,dvs=['total_bill','tip'],group='size')
```
raphaelvallat, line 312: It should be `pg.box_m`

pingouin/multivariate.py
```diff
@@ -254,3 +254,114 @@ def multivariate_ttest(X, Y=None, paired=False):
...
+    References
+    ----------
+    .. [1] Rencher, A. C. (2003). Methods of multivariate analysis (Vol. 492).
+    John Wiley & Sons.
+
+    Examples
+    --------
+    1. Box's M testing 3 covariance matrices from 'tip' dataset
+
+    >>> import numpy as np
```
raphaelvallat, line 308: Please remove unused call to numpy

pingouin/multivariate.py
```diff
@@ -254,3 +254,114 @@ def multivariate_ttest(X, Y=None, paired=False):
...
+        * ``'equal_cov'``: True if ``data`` has equal covariance
+
+    Notes
+    -----
+    This function does not handle missing values. Please
+    handle them (e.g. remove or impute these values)
+    prior to perform the Box's M test.
+    Pooled sample covariance matrix :math:`S_{\\text{pl}}`
+    is calculated as
+    .. math::
```
raphaelvallat, line 290: Please add space before and after the math code block, otherwise it won't be rendered properly in the HTML documentation

pingouin/multivariate.py
```diff
@@ -254,3 +254,114 @@ def multivariate_ttest(X, Y=None, paired=False):
...
+        \\textbf{S}_i}{\\sum_{i=1}^k(n_i-1)},
+    where :math:`n_i` and :math:`S_i` are the sample size and covariance matrix
+    of the :math:`i^{th}` sample, :math:`k` is
+    the number of independent samples.
+    More mathematical expressions can be found in [1]
+
+    References
+    ----------
+    .. [1] Rencher, A. C. (2003). Methods of multivariate analysis (Vol. 492).
+    John Wiley & Sons.
```
raphaelvallat, line 302: Could you also add the reference for using 0.001 as the default significance value? Thanks

pingouin/multivariate.py
```diff
@@ -254,3 +254,114 @@ def multivariate_ttest(X, Y=None, paired=False):
...
+    Examples
+    --------
+    1. Box's M testing 3 covariance matrices from 'tip' dataset
+
+    >>> import numpy as np
+    >>> import pingouin as pg
+    >>> import scipy.stats
+    >>> data = pg.read_dataset('tips')[['total_bill','tip','size']]
+    >>> box_m(data,dvs=['total_bill','tip'],group='size')
+
```
raphaelvallat, line 313: Please remove blank line

pingouin/multivariate.py
```diff
@@ -254,3 +254,114 @@ def multivariate_ttest(X, Y=None, paired=False):
...
+            Chi2	df	    pval	    equal_cov
+    box	45.842377	15.0	0.000056	False
+    """
+
```
raphaelvallat, line 317: Please remove blank line

pingouin/multivariate.py
```diff
@@ -254,3 +254,114 @@ def multivariate_ttest(X, Y=None, paired=False):
...
+        c = - k2
+    else:
+        for idx_cov in range(num_covs):
+            T = - T + (1 / (sizes.iloc[idx_cov] - 1))
+        c = - k1 * (T - (1 / (np.sum(sizes) - num_covs)))
+
+    # calculate U statistics and degree of fredom
+    u = - 2 * (1 - c) * np.log(M)
+    df = 0.5 * num_dvs * (num_dvs + 1) * (num_covs - 1)
+    p = 1 - chi2.cdf(u, df)
```
raphaelvallat, line 363: For consistency with other Pingouin's functions, please use `p = chi2.sf(u, df)`

raphaelvallat (commented):
Just added a few last comments @GanshengT ! Thanks

pingouin/multivariate.py
GanshengT: You mean there should be a blank line before and after the "..math::"? Is there a way that I can see how it renders locally?

GanshengT:
Thank you for the comments. I now know more about python package building.

pingouin/multivariate.py
raphaelvallat: @GanshengT the instructions for building the documentation can be found here: https://pingouin-stats.org/contributing.html

GanshengT:
It seems that the test fails at TestNonParametric.test_harrelldavis

raphaelvallat:
I just checked and the tests that fail are completely unrelated to the current PR, so I can merge safely. Should I merge now or are there any other documentations/tests that you want to add?

GanshengT:
I think this PR is ready to merge.

raphaelvallat:
Merging now ! Thanks again for the PR !

raphaelvallat:
@GanshengT I finally found the time to dive more into this PR, and I've made several improvements to the code (which now uses NumPy broadcasting instead of for loop + pandas), documentation, and unit testing (three different tests against R biotools). You can see the new commit here: https://github.com/raphaelvallat/pingouin/commit/c7ce9abe25501ce187cfe38e687816a2082e51b8

Thanks,
Raphael

PR #177, opened 2021-05-29
Update correlation.py

jinwx:


raphaelvallat:
Hi @jinwx,

What is this PR for? We do want to keep `tail="two-tailed"` as the default in pingouin.corr.

Thanks,
Raphael

sappelhoff:
looks like a correction to me, as the default is currently not `"two-tailed"`:  https://github.com/raphaelvallat/pingouin/blob/c7a074f532aa1e8e7968b08732797e255e8c6f5f/pingouin/correlation.py#L1191

but then perhaps the default should be corrected, and not the docstring?

raphaelvallat:
Oh my bad, you're right I didn't see that the fix was for the distance_corr function! Merging now. Thanks for the PR!

raphaelvallat (approved):

PR #191, opened 2021-08-24
DOCS add how Logistic Regression is modified to match R and statsmodels

discdiver:
The docs didn't say how the scikit-learn default solver was changed. Inquiring minds would have to go to the source code. I added the default solver.


raphaelvallat (approved):

raphaelvallat:
Thank you, merging now!

PR #193, opened 2021-09-02
Only put license name in package metadata

ecederstrand:
The license entry is only supposed to contain the license name, not the full license text. We already have that in the LICENSE file which is included in the package via MANIFEST.in


raphaelvallat (approved):

raphaelvallat:
Thanks for spotting this! Merging now.

Issue #214, opened 2021-11-25
Implemetation of Mardia-Watson-Wheeler test and Moore's Modified Rayleigh test for circular data


J-Wall:
Hi,

I've been looking for a statistics library which implements some tests on circular statistics, or would be happy for me to contribute implementations to.

I'm specifically interested in the Mardia-Watson-Wheeler test for equality of two angular populations (https://www.jstor.org/stable/25049573) and also Moore's Modified Rayleigh test for non-uniformity in weighted angular (vector) data (https://doi.org/10.1093/biomet/67.1.175).

R has an implementation of Mardia-Watson-Wheeler which could be tested against (https://www.rdocumentation.org/packages/circular/versions/0.4-93/topics/watson.wheeler.test). The only implementation of Moore Rayleigh I'm aware of is in Oriana, which is proprietary and closed-source.

Cheers!

H2yx:
excuse me？do u  find a way to implement the Mardia-Watson-Wheeler test for equality of two angular populations？I really need that function!!SOS

raphaelvallat:
@H2yx there was a pull request implementing this test but it was stalled for a long time and I eventually closed it. See the implementation of the test here: https://github.com/raphaelvallat/pingouin/pull/215/files

PR #215, opened 2021-12-01
Moore-rayleigh

J-Wall:
This PR implements the Moore's modified Rayleigh test described in Moore, B. R. (1980). A modification of the Rayleigh test for vector data. Biometrika, 67(1), 175-180. Addresses part of #214.  I've tried to follow your style for code and documentation, and I've included tests based on R* values obtained from the Oriana circular statistics program, and manually calculated p-values.

This test is slightly obscure (in fact, I couldn't find it in any other statistics library...) although it's used a fair bit in animal migration research. The lack of an open-source implementation was bugging me, so I thought I would contribute to an existing library rather than start my own. Cheers!

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -52,6 +52,7 @@ Circular
     circ_mean
     circ_r
     circ_rayleigh
+    circ_moore_rayleigh
     circ_vtest
 
 Contingency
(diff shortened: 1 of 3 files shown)


raphaelvallat:
Hi @J-Wall!

Thanks so much for this. Really clean PR! Two comments after a quick review:

1) For unit testing, can you get higher float precision than 3 decimals? I'm not familiar at all with the Oriana software.

2) Should we combine the circ_moore_rayleigh and circ_rayleigh into a single function? e.g. the Moore's method could be specified as an optional parameter to the main circ_rayleigh function? I'd much prefer this option rather than burdening Pingouin's API with a very rare function.

Thanks,
Raphael

raphaelvallat:
I am closing this PR since there has not been any activity in several months. Feel free to reopen. Thanks

Issue #232, opened 2022-02-08
ttest() error with scipy 1.8.0


adamnarai:
ttest() fails with the following error when using scipy 1.8.0:
```
ImportError: cannot import name '_unequal_var_ttest_denom' from 'scipy.stats.stats'
```
I suspect the problem is that they relocated _unequal_var_ttest_denom() to _stats_py.py
See: https://github.com/scipy/scipy/commit/5a263905fbb61cb4f9a5dfd593451096054373db

raphaelvallat:
Thank you @adamnarai! This should be pretty straightforward to fix. Do you want to submit a quick PR? Otherwise I'll implement this in the next stable release of Pingouin. Note that we do need to keep support for previous versions of scipy for now.

Thanks,
Raphael

adamnarai:
Hi @raphaelvallat,

Sure, I submitted a PR: https://github.com/raphaelvallat/pingouin/pull/234

Best,
Adam

Issue #227, opened 2022-01-24
ttest() TypeError: object of type 'int' has no len()


benpla:
The first sample code fails with:

Code:
```
from pingouin import ttest
x = [5.5, 2.4, 6.8, 9.6, 4.2]
ttest(x, 4).round(2)
```

Every try for ttest farls with the same wessage.
Avova works as expected.
I enter code as simple as possible, but I can not understand the error message.

I uninstalled pingouin using `conda remove pingouin`.
Then installation with pip. Same result.
I installed an older version (`conda install -c conda-forge/label/cf202003 pingouin`).
Same result.

What can be the reason for this?

Here is my output:
```
(py39v1) bp@bplattemac ~ % python
Python 3.9.7 (default, Sep 16 2021, 08:50:36) 
[Clang 10.0.0 ] :: Anaconda, Inc. on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from pingouin import ttest
>>> x = [5.5, 2.4, 6.8, 9.6, 4.2]
>>> ttest(x, 4).round(2)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/bp/opt/anaconda3/envs/py39v1/lib/python3.9/site-packages/pingouin/parametric.py", line 309, in ttest
    stats = pd.DataFrame.from_records(stats, columns=col_order, index=['T-test'])
  File "/Users/bp/opt/anaconda3/envs/py39v1/lib/python3.9/site-packages/pandas/core/frame.py", line 2215, in from_records
    arrays, arr_columns, result_index = maybe_reorder(
  File "/Users/bp/opt/anaconda3/envs/py39v1/lib/python3.9/site-packages/pandas/core/frame.py", line 2165, in maybe_reorder
    length = len(arrays[0])
TypeError: object of type 'int' has no len()
>>> 
```

Pofozaur:
same issue here
```
>>> import numpy as np
>>> import pingouin as pg
>>> np.random.seed(123)
>>> mean, cov, n = [4, 5], [(1, .6), (.6, 1)], 30
>>> x, y = np.random.multivariate_normal(mean, cov, n).T
>>> pg.ttest(x, y)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/kamil/test/venv/lib/python3.8/site-packages/pingouin/parametric.py", line 309, in ttest
[...]

PR #234, opened 2022-02-09
Fixes for dependency (scipy and pandas) updates

adamnarai:
Fixed the bug from https://github.com/raphaelvallat/pingouin/issues/232

To pass unit tests I also:
Implemented your fix @raphaelvallat from https://github.com/raphaelvallat/pingouin/issues/227

DataFrame.apply() does not copy dtypes for some reason ant it is also deprecated (https://pandas.pydata.org/pandas-docs/version/1.4.0/whatsnew/v1.4.0.html), causing unit tests to fail, so I replaced the following line
https://github.com/raphaelvallat/pingouin/blob/dcfdc82bbc7f1ba5991b80717a5ca156617443e8/pingouin/pairwise.py#L461-L462
with
```py
stats = stats.reindex(stats.index.union(idxiter))
```

### DIFF
--- a/pingouin/correlation.py
+++ b/pingouin/correlation.py
@@ -645,7 +645,7 @@ def corr(x, y, alternative='two-sided', method='pearson', **kwargs):
         stats['BF10'] = bayesfactor_pearson(r, n_clean, alternative=alternative)
 
     # Convert to DataFrame
-    stats = pd.DataFrame.from_records(stats, index=[method])
+    stats = pd.DataFrame(stats, index=[method])
 
     # Define order
     col_keep = ['n', 'outliers', 'r', 'CI95%', 'p-val', 'BF10', 'power']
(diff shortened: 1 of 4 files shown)


raphaelvallat:
Hi @adamnarai,

Thank you, this is great! I haven't checked but I am guessing that we should also remove the call to `from_records()` in `correlation.py`: 

https://github.com/raphaelvallat/pingouin/blob/72e447d539f8cfbd6b9c9c5041e267de748fe124/pingouin/correlation.py#L647-L648

and

https://github.com/raphaelvallat/pingouin/blob/72e447d539f8cfbd6b9c9c5041e267de748fe124/pingouin/correlation.py#L865-L866

raphaelvallat (approved):
Good to merge!

raphaelvallat:
Merging now, thanks!

PR #239, opened 2022-02-19
Remove dependency on scipy._morestats

gedeck:
It is clear in this case that labels and title are added to an axis object. The call to `scipy._morestats` can therefore be replaced with direct calls of the matplotlib functions.


raphaelvallat (approved):

raphaelvallat:
Hi @gedeck!

Thank you for the PR! Agreed, I don't remember why I used the call to SciPy in the first place...!

Merging now,
Raphael

Issue #244, opened 2022-02-22
Mediation Analysis: Pass on arguments to logistic_regression


julibeg:
Have been using Pingouin for a while now and it's awesome. However, when I run mediation analyses with binary mediators, `newton-cg` sometimes does not converge and I have to manually change the solver to `sag` or `saga` in the source code (in addition to increasing `max_iter`). I would like to be able to pass on arguments to `sklearn.linear_model.LogisticRegression` in some way (perhaps by adding `logistic_regression_kwargs` to `mediation_analysis` that is then passed on to `_point_estimate` and `logistic_regression`). 
Alternatively, I guess these arguments could also be added to the `options` and read from there.

If you would like to add such functionality, let me know and I'm happy to implement the changes.

PR #245, opened 2022-02-26
optional arguments for logistic_regression() in mediation_analysis()

julibeg:
As discussed in #244.

I added a dict `logreg_kwargs` with the arguments for logistic regression instead of `**kwargs` as this might be cleaner at some later time when you might also want to pass arguments to other functions called by `mediation_analysis()` (e.g. `linear_regression()`). If you prefer the `**kwargs` solution, let me know and I'll change it.

Also, I'm not really familiar with Sphinx, please check if the docstring correct.


raphaelvallat:
Hi @julibeg,

That looks great, thanks! Have you made sure that changing the `logreg_kwargs` change the output of the mediation analysis? Could you perhaps add just one test in the [test_regression.py](https://github.com/raphaelvallat/pingouin/blob/master/pingouin/tests/test_regression.py) file?

Thanks,
Raphael

julibeg:
I made sure by printing the `kwargs` right before `lom.fit(X, y)` in `logistic_regression()` (i.e. [here](https://github.com/raphaelvallat/pingouin/blob/899774de1386e7f038bd8945f6333e4349964f7b/pingouin/regression.py#L872)), but I can also add a test.

julibeg:
OK, I added some more code to test for differences in the actual results.
Something else I noticed: When testing the indirect effect with `assert_almost_equal(ma['coef'][4], 0.0033, decimal=2)`, `2` might not be strict enough for the test to be meaningful, maybe replace it with '3'.

raphaelvallat:
Hi @julibeg,

Thanks! I agree we can use `decimal=3`. However, can you explain what these added lines do:

```python
 with pytest.raises(AssertionError):
    assert_almost_equal(ma['coef'][0], -0.0208, decimal=2)
with pytest.raises(AssertionError):
    assert_almost_equal(ma['coef'][4], 0.0033, decimal=3)
```

Right now, this will test that ma['coef'][0] is **not** equal to -0.0208, and as a result throw an assertion error. Is this the correct behavior?

Thank you
Raphael

julibeg:
```python
# Solve with 0 iterations and make sure that the results are different
ma = mediation_analysis(data=df, x='X', m='Mbin', y='Y', n_boot=2000,
                        logreg_kwargs=dict(max_iter=0))
```
This runs the mediation analysis while not allowing the solver to iterate so that the results are different (however, weirdly not all the coefs are different, just `coef[0]` and `coef[4]`). 
```
with pytest.raises(AssertionError):
    assert_almost_equal(ma['coef'][0], -0.0208, decimal=2)
with pytest.raises(AssertionError):
    assert_almost_equal(ma['coef'][4], 0.0033, decimal=3)

```
This checks that the coefs are indeed different and that the assertion fails.

raphaelvallat:
Got it, thanks! So are we ready to merge?

julibeg:
I think so

raphaelvallat (approved):

Issue #251, opened 2022-03-14
Do eta squared and partial eta squared coincide in one-way repeated measures anova?


Federico2111:
In both the pingouin anova and rm_anova library, it is specified that, in one-way anova, eta squared and partial eta squared coincide: this means that eta squared and partial eta squared coincide both in between- and within-subjects designs, i.e. both in independent and repeated measures designs. I tested the libraries on some data sets and that is indeed true: when I calculated eta squared and partial eta squared on the same data sets, I got the same value, both with the independent and repeated measures anova library.
However, I have been discussing about this with professors, who published about statistics and effect sizes, and say otherwise. Some of them claim, providing demonstrations, that eta squared and partial eta squared coincide only in one-way between-subjects (independent) anova, but NOT in one-way within-subjects (repeated measures) anova. But I have also found publications and websites saying that, in one-way anova, eta squared and partial eta squared always coincide, without specifying the type of design, and thus I suppose including both. 
What's your opinion? 

On a different but related note, I could read in the literature that there is a method to calculate the parametric confidence interval around eta squared, using the non-central F distribution. My suggestion is to add this to the anova libraries. I think it is implemented in some R library. 

Thanks

raphaelvallat:
Hi @Federico2111,

Thanks for opening the issue. I am not a statistican so I fear that I am not the right person to answer this. Pingouin's current behavior is similar to the reference JASP software, as shown below (`n2 == np2 != ng2`)

**Pingouin**

![image](https://user-images.githubusercontent.com/25703792/158428708-f8854f26-dce6-468b-9873-62c47ccdd6cf.png)

**JASP**

![image](https://user-images.githubusercontent.com/25703792/158428580-caf3429f-e1f5-4095-99d1-bc92353a57a4.png)

My preference would therefore **not** to make any modifications to the [pingouin.rm_anova](https://pingouin-stats.org/generated/pingouin.rm_anova.html#pingouin.rm_anova) function, unless we have a very strong rationale to do so. Moreover, what would be the calculation for the new partial eta-squared?

********

Regarding the parametric confidence intervals around the eta-square, I think this is something that could be added to the [pingouin.compute_esci](https://pingouin-stats.org/generated/pingouin.compute_esci.html#pingouin.compute_esci) function. I would suggest opening a new issue specifically for this. Please include some examples from other statistical softwares as well as any relevant literature.

Thank you,
Raphael

Federico2111:
Thanks @raphaelvallat,

[...]

PR #254, opened 2022-03-19
Fix eta-squared calculation in repeated measures ANOVA design

raphaelvallat:
This PR fixes the issue described in https://github.com/raphaelvallat/pingouin/issues/251

- In one-way ANOVA design, the calculation of the eta-squared `n2` is now the same as the generalized eta-squared `ng2`. Previous versions of Pingouin used the same formula for the eta-squared and the partial eta-squared `np2`, which is invalid. Note that this invalid behavior is currently the default in JASP, but not in JAMOVI.

- The default effect size metric in repeated measures ANOVA design is now the generalized eta-squared `ng2`.

- The calculation of the eta-squared `n2` in two-way repeated measures ANOVA design was invalid (approximatively similar to the partial eta-squared). This has been fixed.

- All effect sizes (`n2`, `np2` and `ng2`) have now been tested against JAMOVI, for both one-way and two-way repeated measures design.

I am looking for one or two reviewers to review this PR. 

Thank you,
Raphael


raphaelvallat:
Still looking for 1-2 reviewers to review this PR

PR #261, opened 2022-04-10
fix the document error

xfz329:
i found the document error using pingouin.
in fact i found sevearl this  kind of errors using pingouin. i will figure them out the later.


raphaelvallat:
Thanks for catching this! Merging now

raphaelvallat (approved):

Issue #272, opened 2022-06-13
LogisticRegression fails when X is a single predictor


raphaelvallat:
See https://github.com/scikit-learn/scikit-learn/issues/23605

This leads to unit tests failure in Pingouin.

The easiest fix is to downgrade scikit-learn to version 1.0:

```shell
pip install scikit-learn==1.0
```

PR #278, opened 2022-06-18
Use scikit-learn<1.1.0

raphaelvallat:
Temporary fix for #272 

We will need to revert these changes when the next stable version of scikit-learn is released (1.1.2): https://github.com/scikit-learn/scikit-learn/pull/23608


poldrack:
it appears that the problem mentioned above was fixed in sklearn 1.1.2 - can this requirement be updated now?

raphaelvallat:
Hi @poldrack — of course, please see https://github.com/raphaelvallat/pingouin/pull/300

However, I am not planning to release a new version of Pingouin before at least 1-2 months.

Thanks,
Raphael

poldrack:
gotcha, thanks!

PR #290, opened 2022-07-06
Add significance star option to ttest output

1nathanliang:
To quickly reference the results of a pingouin statistical test like the t-test (I've only changed this one as an example), it might be helpful to include significance stars or a null effect label in the output data frame

### DIFF
--- a/pingouin/parametric.py
+++ b/pingouin/parametric.py
@@ -17,7 +17,7 @@ from pingouin import (
 __all__ = ["ttest", "rm_anova", "anova", "welch_anova", "mixed_anova", "ancova"]
 
 
-def ttest(x, y, paired=False, alternative="two-sided", correction="auto", r=0.707, confidence=0.95):
+def ttest(x, y, paired=False, alternative="two-sided", correction="auto", r=0.707, confidence=0.95, stars=True):
     """T-test.
 
     Parameters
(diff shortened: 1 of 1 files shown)


raphaelvallat:
Hi @1nathanliang,

Thanks for opening the PR. I am a bit reluctant to include significance stars in the output of the pingouin.ttest function -- or any Pingouin function actually. I like to keep the output as factual as possible and leave the higher-level interpretation to the users.

However, as a compromise, I have been thinking about implementing a T-test equivalent to the existing pingouin.rcorr function (see below), which would perform a T-test (paired or unpaired) instead of a correlation. In other words, you would see the T values on the lower half of the matrix and the p-values (stars) on the upper half. This function could be used as a quick replacement for pingouin.ttest in situations where users mostly care about the T and p-values.

image

Let me know what you (and others) think,

Thanks
Raphael

remrama:
I agree with the general reluctance. They're not for everybody, plus it's trivial to extract stars in one's own custom line, eg,
```python
df["stars"] = df["p-val"].map(lambda p: "*" * sum([ p<cutoff for cutoff in [.05, .01, .001] ]))
```
I think you offer a great compromise @raphaelvallat , stars feel acceptable there since they're only filling empty space (unless you wanted to add effect size or something in the upper triangle, which might also be nice). Either way, the new function would be a fantastic feature!

1nathanliang:
@raphaelvallat Makes sense, and thanks for your work on this great package :) and @remrama agreed, I have a longer version of what you've written—thanks for the concise recommendation!

PR #292, opened 2022-07-18
Fix numerical stability issue in multivariate_normality

gkanwar:
For large n, the lognormal variance was given by a function of log(1+x) for x very close to zero. Numpy provides the more stable function log1p(x) to address cases where 1+x = 1 in floating point representations, which would otherwise incorrectly yield log(1+x) = 0.

I added a test to check this case. The previous version of the code failed this check (giving p = nan).


raphaelvallat (approved):
Awesome, thanks for the PR. Merging now!

PR #299, opened 2022-08-26
Handle single-sample comparsion in pairwise_test

George3d6:
Due to a scipy interface change the pairwise_test can fail when obtaining the t-value from a single sample comparison, this, I believe, fixes that issue.

### DIFF
--- a/pingouin/multivariate.py
+++ b/pingouin/multivariate.py
@@ -58,7 +58,7 @@ def multivariate_normality(X, alpha=0.05):
     >>> data = pg.read_dataset('multivariate')
     >>> X = data[['Fever', 'Pressure', 'Aches']]
     >>> pg.multivariate_normality(X, alpha=.05)
-    HZResults(hz=0.540086101851555, pval=0.7173686509622385, normal=True)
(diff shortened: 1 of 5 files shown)


raphaelvallat:
Hi @George3d6,

Thank you for this!

1. Can you be more specific about which versions of SciPy have this potential bug? Did it appear in SciPy 1.9?
2. Does this issue only occur for one-sample T-tests?
3. There seems to be an unrelated unit test failure, I'll fix that now

George3d6:
I'm actually unsure when the scipy interface might have changed, or if this was always the case, I'll look into it.
But as can bee seen, the current scipy interface is returning a tuple that the code treats as expecting to be a float or int, so the impl is most certainly wrong

George3d6:
@raphaelvallat as far as I can see this isn't a recent change in the scipy interface as I suspected, so I guess the bug was always there.

raphaelvallat (approved):

raphaelvallat:
Got it, well I think we're good to merge. Thanks again for the PR!

Issue #306, opened 2022-10-10
Support for passing most plotting functions to seaborn grids


remrama:
Hi @raphaelvallat , I'm wondering how tied you are to generating figures explicitly in most of pingouin's plotting functions (e.g., `plot_blandaltman`). Currently, most of these plotting functions generate a figure explicitly when the `ax` parameter is not provided, something like:
```python
if ax is None:
    fig, ax = plt.subplots(1, 1, figsize=figsize, dpi=dpi)
```

I think it might be better or more flexible to generate a figure with default size the way seaborn does:
```python
if ax is None:
    ax = plt.gca()
```

Seems trivial, but if it's set up this way, then the function can be used to "map" plots onto seaborn grids (e.g., [this demo](http://seaborn.pydata.org/examples/many_facets.html)). It would be great if pingouin's plotting functions could be applied to this popular seaborn feature. Other very minor changes would also need to happen, namely taking a `color` parameter (see [here](https://seaborn.pydata.org/generated/seaborn.FacetGrid.map.html#seaborn.FacetGrid.map:~:text=A%20plotting%20function%20that%20takes%20data%20and%20keyword%20arguments.%20It%20must%20plot%20to%20the%20currently%20active%20matplotlib%20Axes%20and%20take%20a%20color%20keyword%20argument.%20If%20faceting%20on%20the%20hue%20dimension%2C%20it%20must%20also%20take%20a%20label%20keyword%20argument.)).

Other reasons to favor this change:
- As far as I can tell, you don't lose anything except the ability to specify figure details. BUT that's where the `ax` parameter gets its use; if someone wants a figure a specific size, they open up the figure themselves (specifiying size, dpi, etc.) and then pass the ax in. Ultimately this has more complete flexibility and suggests the user build their own figure if they are concerned about aesthetic details.
- Currently, when `ax` is used, `figsize` (and and sometimes `dpi`) is ignored. This is slightly confusing so this avoids that as well.

This would of course _not_ apply to any figures that need to build their own, like anything using seaborn facetgrids. At a glance, I think this change could apply to the following functions:
- `plot_blandaltman`
- `plot_paired`
- `plot_circmean`
- `qqplot` (also dropping the automatic Title here would help compatibility with seaborn)

Note there is a 3rd option where you could have both. You could check if a figure already exists, and then if not build one with the arguments passed to the pingouin plotting function. I don't favor this much, but just mentioning it.

If you're into this I'd be happy to make the pull request.

[...]

PR #314, opened 2022-11-25
Plotting seaborn.FacetGrid compatibility

remrama:
This addresses a feature request in #306 to offer more flexibility in pingouin's plotting module. The changes were targeted at making some of the functions useable with the `g.map_dataframe()` method attached to seaborn grids, but they offer more flexibility when used alone as well.

The main things:
* Removed all fig and dpi keyword arguments (users should be advised to generate their own figures first, eg with `plt.subplot` and then pass the `ax` to pingouin)
* Removed the `plot_blandaltmann` keyword argument `scatter_kws` and now any `kwarg` gets passed through. This was done because `gmap_dataframe` requires the passing of a `color` keyword (and `label`), and it can be useful at times.
* Note that now, without passing `fig` through, the default axis that pops up is not always square. So for the few plotting functions that probably _should_ be square (qqplot, circmean), I added an `ax.set_aspect("equal")` in the docstring examples.

If you want to see some arbitrary examples of the use-case I'm talking about:

```python
import pingouin as pg
import seaborn as sns
import matplotlib.pyplot as plt

iris = sns.load_dataset("iris")

g = sns.FacetGrid(iris, col="species", aspect=1)
g.map(pg.qqplot, "petal_width")
plt.show()

g = sns.FacetGrid(iris, col="species", hue="species", sharey=False)
g.map(pg.plot_blandaltman, "petal_width", "sepal_width")
plt.show()
```


raphaelvallat:
Hi @remrama,

Thanks for the PR, great idea! A few comments:

- [ ] Can you please update the unit tests?
- [ ] Update the changelog
- [ ] Does this require to bump seaborn's version in the requirements/setup?
- [ ] I do feel that`ax.set_aspect("equal")` should be the default for qqplot and circmean. Is there a way that we could ensure that, e.g. by including a `square=True` parameter like in seaborn.heatmap?

raphaelvallat:
@remrama I've rebased the PR to master, make sure you `git pull` before making new changes!

remrama:
@raphaelvallat I think this is good to go?

It passes unittests, changelog is updated, and `square` keyword arguments were added to `qqplot()` and `plot_circmean()`. Nothing to worry about for seaborn version requirements. `map()` was present in version 0.10, and you already require 0.11.

raphaelvallat (approved):

raphaelvallat:
@remrama just made some very minor edits to documentation. Merging when the CIs are complete

PR #315, opened 2022-11-29
Updating of shift plot in plotting.py

turkalpmd:
The entire library is absolutely amazing; thank you for developing it. I came up with a minor improvement for the "Shift plot". No matter what data array we provide, this graphic shows X and Y. There may be disagreement about its use in academic papers. I believed that by defining X and Y, this conflict would be addressed. I created an edit for this where the user can configure the X and Y variables as well as the plot title. The labels being on the Y axis creates problems with the size of the figure, thus I located them as a legend in the plot.

I have ideas for other modules and would like to add them when I have the time. My initial contribution would be adding the odds ratio and CI-range values to the chi-sqr module.

Here my test code;

[plotting_test.zip](https://github.com/raphaelvallat/pingouin/files/10117052/plotting_test.zip)

### DIFF
--- a/pingouin/plotting.py
+++ b/pingouin/plotting.py
@@ -657,25 +657,27 @@ def plot_paired(
 
 
 def plot_shift(
-    x,
-    y,
-    paired=False,
-    n_boot=1000,
-    percentiles=np.arange(10, 100, 10),
-    confidence=0.95,
-    seed=None,
-    show_median=True,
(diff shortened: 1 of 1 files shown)


raphaelvallat:
Closing this issue which is a duplicate of https://github.com/raphaelvallat/pingouin/pull/316

PR #316, opened 2022-11-30
Add axis title to plot_shift

turkalpmd:
The entire library is absolutely amazing; thank you for developing it. I came up with a minor improvement for the "Shift plot". No matter what data array we provide, this graphic shows X and Y. There may be disagreement about its use in academic papers. I believed that by defining X and Y, this conflict would be addressed. I created an edit for this where the user can configure the X and Y variables as well as the plot title. The labels being on the Y axis creates problems with the size of the figure, thus I located them as a legend in the plot.

I have ideas for other modules and would like to add them when I have the time. My initial contribution would be adding the odds ratio and CI-range values to the chi-sqr module.

Here my test code;
[plotting_test.zip](https://github.com/raphaelvallat/pingouin/files/10122374/plotting_test.zip)

### DIFF
--- a/pingouin/plotting.py
+++ b/pingouin/plotting.py
@@ -657,18 +657,18 @@ def plot_paired(
 
 
 def plot_shift(
-                x,
-                y,
-                paired=False,
-                n_boot=1000,
-                percentiles=np.arange(10, 100, 10),
-                confidence=0.95,
-                seed=None,
-                show_median=True,
-                violin=True,
-                title="Plot Shift",
-                x_label="X",
-                y_label="Y"                
+    x,
+    y,
+    paired=False,
+    n_boot=1000,
+    percentiles=np.arange(10, 100, 10),
+    confidence=0.95,
+    seed=None,
+    show_median=True,
+    violin=True,
+    title="Plot Shift",
+    x_label="X",
+    y_label="Y",
 ):
 
     """Shift plot.
@@ -679,7 +679,7 @@ def plot_shift(
         First and second set of observations.
     paired : bool
         Specify whether ``x`` and ``y`` are related (i.e. repeated measures) or independent.
-    
+
     .. versionadded:: 0.3.0
     n_boot : int
         Number of bootstrap iterations. The higher, the better, the slower.
@@ -695,43 +695,43 @@ def plot_shift(
     violin: boolean
         If True (default), plot the density of X and Y distributions. Defaut set to True.
     title: str
-        Define which title should be used for all plot. 
+        Define which title should be used for all plot.
             Default set "Plot Shift"
     x_label: str
-        Define which label should be used for X array. 
+        Define which label should be used for X array.
             Default set "X"
     y_label: str
-        Define which label should be used for Y array.  
-            Default set "Y"        
-    
+        Define which label should be used for Y array.
+            Default set "Y"
+
     Returns
     -------
     fig : matplotlib Figure instance
         Matplotlib Figure. To get the individual axes, use fig.axes.
-    
+
     See also
     --------
     harrelldavis
-    
+
     Notes
     -----
     The shift plot is described in [1]_. It computes a shift function [2]_ for two (in)dependent
     groups using the robust Harrell-Davis quantile estimator in conjunction with bias-corrected
     bootstrap confidence intervals.
-    
+
     References
     ----------
     .. [1] Rousselet, G. A., Pernet, C. R. and Wilcox, R. R. (2017). Beyond
            differences in means: robust graphical methods to compare two groups
            in neuroscience. Eur J Neurosci, 46: 1738-1748.
            doi:10.1111/ejn.13610
-    
+
     .. [2] https://garstats.wordpress.com/2016/07/12/shift-function/
-    
+
     Examples
     --------
     Default shift plot
-    
+
     .. plot::
         >>> import numpy as np
         >>> import pingouin as pg
@@ -739,11 +739,11 @@ def plot_shift(
         >>> x = np.random.normal(5.5, 2, 50)
         >>> y = np.random.normal(6, 1.5, 50)
         >>> fig = pg.plot_shift(x, y)
-    
+
     With different options
-    
+
     .. plot::
-    
+
         >>> import numpy as np
         >>> import pingouin as pg
         >>> np.random.seed(42)
@@ -777,7 +777,7 @@ def plot_shift(
     delta = y_per - x_per
 
     # Compute bootstrap distribution of differences
-    rng = np.random.RandomState(seed) 
+    rng = np.random.RandomState(seed)
     if paired:
         bootsam = rng.choice(np.arange(nx), size=(nx, n_boot), replace=True)
         bootstat = hd(y[bootsam], percentiles, axis=0) - hd(x[bootsam], percentiles, axis=0)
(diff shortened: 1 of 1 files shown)


remrama:
I'm all about improving pingouin's plotting capabilities, so I love this idea! But is there a reason that changing axis labels and adding a legend should be done _within_ pingouin, rather than having the user modify those things after setting the base figure with pingouin? If it's not critical to have this built-in, then I think it should be more optional as a post-pingouin modification (for a variety of reasons).

turkalpmd:
Yes, I  agree with you. The library is currently in excellent shape.

Q: But is there a reason that changing axis labels and adding a legend
should be done within pingouin, rather than having the user modify those
things after setting the base figure with pingouin?

A: Yes, since the plt.something() modules I can configure after plotting
only support ax2 plot. I cannot configure to ax1 plot.

I did not understand your suggestion as post-pingouin modification, sorry.

My solution:
These variables can stay precisely how they are. For instance, the shift
plot function will continue to function as before if the variables I
recommend are not provided. My suggestions are configurable. In order to
create the plots for my thesis, I thought that I should write what X and Y
are in legend. I started by switching the axes. Short-length labels were
not a problem. The figure in ax1 experienced a significant display issue
when I attempted to use long-length labels, while the figure in ax2
disappeared. Then, I believed it made sense to include X and Y as minor
characters in the plot as a legend. Additionally, I considered writing what
I'm comparing to the figure as the title, this approach makes sense to me.

Finally, I really appreciate your contribution. You are a fantastic
alternative for me because I hated SPSS. I'll use the shift_plot function
exactly how I designed it. Perhaps others also will find this more helpful.
I also have ideas for other components, and I want to talk to you about
them when I have time.


remrama ***@***.***>, 30 Kas 2022 Çar, 23:33 tarihinde şunu
yazdı:

remrama:
Ok I see. @turkalpmd you should be able to access both axes with pingouin's current implementation by extracting them from the returned `fig` variable. So to extend the example from pingouin's docs:

```python
import numpy as np
import pingouin as pg
np.random.seed(42)
x = np.random.normal(5.5, 2, 50)
y = np.random.normal(6, 1.5, 50)
fig = pg.plot_shift(x, y)
ax1, ax2 = fig  # where ax1 is on top and ax2 is on bottom
ax2.set_xlabel("New x label on bottom axis")
ax2.set_ylabel("New y label on bottom axis")
ax2.legend(title="Legend for bottom axis")
```

Note that even if figure was not returned, you can always extract it like this:
```python
fig = plt.figure()
# and then access all available axes from that figure with
fig.axes
```

As far as I can tell this should solve all the issues you were looking to resolve with this PR, is that correct? Or did I miss something else?

remrama:
**edited `plt.figure()` --> `plt.gcf()` in the example code

turkalpmd:
When I run, I get this error;

5 y = np.random.normal(6, 1.5, 50) 7 fig = pg.plot_shift(x, y) ----> 9 ax1,
ax2 = fig # where ax1 is on top and ax2 is on bottom 10 ax2.set_xlabel("New
x label on bottom axis") 11 ax2.set_ylabel("New y label on bottom axis")
TypeError: cannot unpack non-iterable Figure object
Okay, no problem for me. I simply wanted to share my improvement on code. I
will use this code, because easier for me.  Please keep in mind that not
everyone is well proficient with the Matplotlib library.

remrama ***@***.***>, 1 Ara 2022 Per, 00:17 tarihinde şunu
yazdı:

remrama:
Sorry! I had another typo. I edited the original post, but basically the line to extract axes should have read like this:
```python
ax1, ax2 = fig.axes
```

Another option would be to access them directly from the `fig.axes` list:
```python
fig.axes[1].set_xlabel("New x label")
```

The moral of the story is that you can use `fig.axes` to access a list of all the axes on the figure, and from there you can implement any modifications (e.g., axis labels, legends).

raphaelvallat:
Hi @turkalpmd,

Thanks for opening the PR, and for your kind words of Pingouin — really appreciate it! Two things:

1. The PR is failing (lint issue) because I think you are using a code editor that automatically changed some of the formatting of Pingouin's code. You can solve the lint issue by doing:

```bash
pip install black
cd pingouin/pingouin
black . --line-length=100
```

2. As for the PR in itself, I do agree with @remrama that this is something that can be changed _a posteriori_ quite easily and I would suggest we just add an example to the documentation of the function instead of modifying the code. Alternatively, we can allow `x` and `y` to be pandas.Series, in which case the labels would be automatically extracted from the name of the series.

raphaelvallat:
Note for self: this issue is a duplicate of https://github.com/raphaelvallat/pingouin/pull/316

turkalpmd:
Thank you for all. I can take the same result with this code;

Maybe these directions are should be mentioned in
https://pingouin-stats.org/generated/pingouin.plot_shift.html#pingouin.plot_shift

import numpy as np
import pingouin as pg
from matplotlib.legend_handler import HandlerPatch
import matplotlib.patches as mpatches

np.random.seed(42)
x = np.random.normal(5.5, 2, 50)
y = np.random.normal(6, 1.5, 50)

fig = pg.plot_shift(x, y)

#ax1, ax2 = fig.axes # where ax1 is on top and ax2 is on bottom
fig.axes[0].set_xlabel("New x label")
fig.axes[0].set_ylabel("New y label")
colors = ["#88bedc", "#cfcfcf"]

texts = ["X","Y"]

from matplotlib.legend_handler import HandlerPatch
class HandlerEllipse(HandlerPatch):
def create_artists(self, legend, orig_handle,
xdescent, ydescent, width, height, fontsize, trans):
center = 0.5 * width - 0.5 * xdescent, 0.5 * height - 0.5 * ydescent
p = mpatches.Ellipse(xy=center,
width=height + xdescent,
height=height + ydescent)
self.update_prop(p, orig_handle, legend)
p.set_transform(trans)
return [p]


c = [mpatches.Circle((0.1, 0.1), radius = 0.25, facecolor=colors[i],
edgecolor="none" ) for i in range(len(texts))]

fig.axes[0].legend(c,
texts,
bbox_to_anchor=(1, 0.2),
loc='best',
ncol=1,
prop = {'size' : 8},
handler_map={mpatches.Circle: HandlerEllipse()}).get_frame()


fig.axes[0].set_title("New title")
plt.show()

remrama ***@***.***>, 1 Ara 2022 Per, 00:42 tarihinde şunu
yazdı:

turkalpmd:
> Hi @turkalpmd,
> 
> Thanks for opening the PR, and for your kind words of Pingouin — really appreciate it! Two things:
> 
> 1. The PR is failing (lint issue) because I think you are using a code editor that automatically changed some of the formatting of Pingouin's code. You can solve the lint issue by doing:
> 
> ```shell
> pip install black
> cd pingouin/pingouin
> black . --line-length=100
> ```
> 
> 2. As for the PR in itself, I do agree with @remrama that this is something that can be changed _a posteriori_ quite easily and I would suggest we just add an example to the documentation of the function instead of modifying the code. Alternatively, we can allow `x` and `y` to be pandas.Series, in which case the labels would be automatically extracted from the name of the series.

Yes, I believe Matplotlib can now address this problem. I can publish the relevant code on Kaggle or elsewhere if you like.

remrama:
Great! I agree that an example like this in the documentation would be nice. It might prevent new users from going through the same process and make the shift plot more accessible.

turkalpmd:
> Great! I agree that an example like this in the documentation would be nice. It might prevent new users from going through the same process and make the shift plot more accessible.

If you could tell me where I should submit this?

remrama:
See the [section here](https://pingouin-stats.org/contributing.html#checking-and-building-documentation) about contributing to pingouin, specifically about contributing to the documentation.

I'm thinking you could add a 3rd example to the already 2 on the [`plot_shift` doc page](https://pingouin-stats.org/generated/pingouin.plot_shift.html#pingouin.plot_shift). Or, even simpler, you could make a very small modification to the 2nd example that just renames the axis labels as you've done. Even without the legend, just showing that `fig.axes[0].set_xlabel` works might be enough to tip users off to the fact that you can mod things that way.

turkalpmd:
I implemented your suggestions and submitted a pull request for review.

remrama ***@***.***>, 1 Ara 2022 Per, 02:19 tarihinde şunu
yazdı:

PR #317, opened 2022-12-05
Shift plot figure update

turkalpmd:
I added Matplotlib directions for the second plot_shift example figure. [Here](https://pingouin-stats.org/generated/pingouin.plot_shift.html#pingouin.plot_shift)


turkalpmd:
@raphaelvallat and @remrama is it okay?

remrama:
For me, editing docs can be really confusing because there is a difference between the docstrings of the functions themselves and the auto-generated html output. I think something happened here where all the edits are in the html output instead of the actual python docstring which is where they would need to be made. (See Pingoun's [contributing docs](https://pingouin-stats.org/contributing.html#checking-and-building-documentation) section.) As far as I can tell, the current PR does not actually change the relevant code that would be necessary to update the website docs.

You would need to edit this section:
https://github.com/raphaelvallat/pingouin/blob/c97b0ef0bf77d984eb2cb1efc5f133d3b4e7ebde/pingouin/plotting.py#L733-L741

@turkalpmd if what you're really after is to just have this included in the docs, I'm happy to submit a new PR that takes care of it! But if you are trying to work through this (or any other PR) I would suggest starting by submitting an Issue first. That's where all the details/decisions should really be worked out before attacking the PR. This is also [stated](https://pingouin-stats.org/contributing.html#code-guidelines) in Pingouin's contributing guidelines. Let me know what you'd prefer, but either way I'd suggest closing this PR for now.

turkalpmd:
I misunderstood the sphinx syntax, and I'm editing it this way.

PR #332, opened 2023-01-02
pairwise non parametrics description

turkalpmd:
I implemented IQR, but if necessary, I can also implement IQR-I and IQR-3.


turkalpmd:
I think is here is asserting the modification.

https://github.com/raphaelvallat/pingouin/blob/6f4a7b483e4abb320148880bb507d96fb98f271d/pingouin/tests/test_pairwise.py#L191

PR #341, opened 2023-02-16
fix: deprecated parameter

bitsnaps:
The only thing need to be changed is `errorbar` instead of `ci` argument in `pointplot` function, which is now deprecated.


raphaelvallat:
Thanks for the PR! For reference, this deprecation has been introduced in seaborn [0.12.0](https://seaborn.pydata.org/whatsnew/v0.12.0.html) (September 2022).

raphaelvallat (approved):

PR #342, opened 2023-02-20
clip r values

### LINKED ISSUE #340: AssertionError for bicor when n == 2
Hi,

I would like to use the bicor method within pairwise_corr. My data contains a lot of nan values that I would like to exclude with nan_policy='pairwise'. Unfortunately, there are also cases where the overlap of notnan values is only 2. In that case I would like to get nan for correlation (or anything else, I don't care too much...), but instead I am getting this AssertionError (that I am btw not getting when replacing the 4 with a 3...):
```
pingouin.corr([1, np.nan, 2, 4], [1, 2, np.nan, 3], method='bicor')
```
```
---------------------------------------------------------------------------
AssertionError                            Traceback (most recent call last)
Input In [55], in <cell line: 1>()
----> 1 pingouin.corr([1, np.nan, 2, 4], [1, 2, np.nan, 3], method='bicor')

File ~/miniconda3/envs/glue/lib/python3.8/site-packages/pingouin/correlation.py:648, in corr(x, y, alternative, method, **kwargs)
    644 else:
    645     ci = compute_esci(
    646         stat=r, nx=n_clean, ny=n_clean, eftype="r", decimals=6, alternative=alternative
    647     )
--> 648     pr = power_corr(r=r, n=n_clean, power=None, alpha=0.05, alternative=alternative)
    650 # Recompute p-value if tail is one-sided
    651 if alternative != "two-sided":

File ~/miniconda3/envs/glue/lib/python3.8/site-packages/pingouin/power.py:852, in power_corr(r, n, power, alpha, alternative)
    845 assert alternative in [
    846     "two-sided",
    847     "greater",
    848     "less",
    849 ], "Alternative must be one of 'two-sided' (default), 'greater' or 'less'."
    851 if r is not None:
--> 852     assert -1 <= r <= 1
    853     if alternative == "two-sided":
    854         r = abs(r)

AssertionError: 
```

remrama:
Fixes #340 

I added `np.clip` to `pg.corr`, right after where all the `r` values are calculated and before being passed to `power_corr`, so I think this should catch all similar cases. Rather than _always_ clipping, I only do it if the r value exceeds -1 or 1 _and_ `np.isclose` to abs(1). I don't imagine getting r values exceeding those bounds, but this felt safer to prevent accidentally clipping  something super-high down to 1.

Re-running the original code that produced the error will work now:

```python
import numpy as np
import pingouin as pg
pg.corr([1, np.nan, 2, 4], [1, 2, np.nan, 3], method='bicor')
# bicor  2  1.0  [1.0, 1.0]    NaN      1
```

### DIFF
--- a/pingouin/correlation.py
+++ b/pingouin/correlation.py
@@ -637,6 +637,10 @@ def corr(x, y, alternative="two-sided", method="pearson", **kwargs):
     n_outliers = sum(outliers) if "outliers" in locals() else 0
     n_clean = n - n_outliers
 
+    # Rounding errors caused an r value marginally beyond 1
+    if abs(r) > 1 and np.isclose(abs(r), 1):
+        r = np.clip(r, -1, 1)
+
     # Compute the parametric 95% confidence interval and power
     if abs(r) == 1:
         ci = [r, r]


raphaelvallat (approved):
LGTM! Unrelated CI error that I need to fix soon. Thanks for the PR!

PR #344, opened 2023-02-22
hotfix: CI crash in test_power_chi2 [WIP]

raphaelvallat:


raphaelvallat:
The range on which the chi-square effect size was evaluated was unecessary too large and caused a Python crash in latest version of SciPy

musicinmybrain:
I spotted this failure in Fedora 39/Rawhide (and noted that it will soon affect Fedora 38/Branched). I can confirm that applying this PR as a downstream patch to 0.5.3 fixes the problem in Fedora, too.

PR #347, opened 2023-03-01
Add check for noncentrality parameters.

### LINKED ISSUE #343: Crash when calculating power with too large effect size
This crash occurs for me with the following code, although it may not work identically on all platforms:
```python
import numpy as np
from pingouin.effsize import compute_effsize
from pingouin.power import power_ttest

x = np.ones(10) * 0.3
y = np.ones(10) * 0.6
d = compute_effsize(x, y, paired=True)
print(d)
print(power_ttest(d, x.size))
```
Running this code results in the output
```
-3242591731706756.5
terminate called after throwing an instance of 'boost::wrapexcept<boost::math::rounding_error>'
  what():  Error in function boost::math::itrunc<double>(double): Value 2.3657402561699292e+31 can not be represented in the target integer type.
Aborted (core dumped)
```

The root cause of this is related to https://github.com/scipy/scipy/issues/17916. However, it occurs in `power_ttest()` due to calling `scipy.stats.nct.cdf()` with the huge value of `nc` based on `d`:
https://github.com/raphaelvallat/pingouin/blob/c66b6853cfcbe1d6d9702c87c09050594b4cacb4/pingouin/power.py#L146-L168

Perhaps a check should be done in `compute_effsize()` to check whether the variance to too small? Or check whether the effect size magnitude is too high in `power_ttest()`?

agkphysics:
Fixes #343, caused by scipy/scipy#17916.

### DIFF
--- a/pingouin/power.py
+++ b/pingouin/power.py
@@ -15,6 +15,22 @@ __all__ = [
 ]
 
 
+def _check_nc(dist, nc):
+    """Check if non-centrality parameter is too large for the given distribution.
+    This is a workaround for scipy/scipy#17916 which can hopefully be removed at some point.
+    """
+    if dist is stats.ncx2 and nc >= float(2**32 - 1):
+        warnings.warn("Non-centrality parameter is too large for the ncx2 distribution.")
+        return False
+    if dist is stats.ncf and nc >= float(2**32):
+        warnings.warn("Non-centrality parameter is too large for the ncf distribution.")
+        return False
+    if dist is stats.nct and (nc <= float(-(2**16)) or nc >= float(2**16)):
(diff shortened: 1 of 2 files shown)


agkphysics:
I notice that coverage fails because of the `greater` and `less` paths. Do you want me to add these extra test cases for `ttest` and `ttest2n`?

raphaelvallat:
Hi @agkphysics,

This is great, thank you! Yes can you please add the extra test? 

Btw I see that https://github.com/scipy/scipy/issues/17916 has now been fixed and should be included in scipy 1.11.0.

agkphysics:
@raphaelvallat I have added the additional tests now.

raphaelvallat (approved):
Approved! Thanks so much

Issue #349, opened 2023-03-06
Variablename "C" leads to a crash in pg.plot_rm_corr


caiusno1:
The following code crashes with a not understandable error message:

```python
import pandas as pd
import pingouin as pg

df_test = pd.DataFrame({"C":[1,2,3,4,5,6], "X":[1,2,3,4,5,6], "VP":[1,1,2,2,3,3]})
pg.plot_rm_corr(df_test, x="C", y="X", subject="VP")
```
Trace
```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
File /opt/conda/lib/python3.10/site-packages/patsy/compat.py:36, in call_and_wrap_exc(msg, origin, f, *args, **kwargs)
     35 try:
---> 36     return f(*args, **kwargs)
     37 except Exception as e:

File /opt/conda/lib/python3.10/site-packages/patsy/eval.py:169, in EvalEnvironment.eval(self, expr, source_name, inner_namespace)
    168 code = compile(expr, source_name, "eval", self.flags, False)
--> 169 return eval(code, {}, VarLookupDict([inner_namespace]
    170                                     + self._namespaces))

File <string>:1, in <module>

TypeError: 'Series' object is not callable

The above exception was the direct cause of the following exception:

PatsyError                                Traceback (most recent call last)
Input In [124], in <cell line: 2>()
      1 df_test = pd.DataFrame({"C":[1,2,3,4,5,6], "X":[1,2,3,4,5,6], "VP":[1,1,2,2,3,3]})
----> 2 pg.plot_rm_corr(df_test, x="C", y="X", subject="VP")

File /opt/conda/lib/python3.10/site-packages/pingouin/plotting.py:1005, in plot_rm_corr(data, x, y, subject, legend, kwargs_facetgrid, kwargs_line, kwargs_scatter)
    996 # Calculate rm_corr
    997 # rmc = pg.rm_corr(data=data, x=x, y=y, subject=subject)
    998 
   (...)
   1002 # Q allows to quote variable that do not meet Python variable name rule
   1003 # e.g. if variable is "weight.in.kg" or "2A"
   1004 formula = "Q('%s') ~ C(Q('%s')) + Q('%s')" % (y, subject, x)
-> 1005 model = ols(formula, data=data).fit()
   1007 # Fitted values
   1008 data["pred"] = model.fittedvalues

[...]

PR #351, opened 2023-03-08
hotfix: plot_rm_corr crash with specific column names

remrama:
Should address #349 

I added the assertion checks right above the use of `ols` as opposed to start of each function, only because it seemed easier for a reader to extract the motivation for them. lmk if you prefer them at the top.


raphaelvallat (approved):
Thanks so much!

PR #352, opened 2023-03-21
added kwarg passing to pairwise_tests

DavidALloyd:
Added ability to pass kwargs to downstream scipy functions using pairwise_tests. Previously did not have that ability, so changing things like the zero method or confidence weren't possible through pairwise_tests. pg.wilcoxon and pg.mwu already pass **kwargs downstream, so I simply modified pairwise_tests to accept **kwargs as a parameter and pass it as an argument when Wilcoxon, MWU, or ttest are called. Tested on example data and works without issue.

### DIFF
--- a/pingouin/pairwise.py
+++ b/pingouin/pairwise.py
@@ -47,6 +47,7 @@ def pairwise_tests(
     return_desc=False,
     interaction=True,
     within_first=True,
+    **kwargs
 ):
     """Pairwise tests.
 
@@ -133,6 +134,9 @@ def pairwise_tests(
         and between * within otherwise.
 
         .. versionadded:: 0.3.6
+    **kwargs : optional
+    Optional argument(s) passed to the lower-level pingouin functions, such as pg.wilcoxon
+    or pg.mwu
 
     Returns
     -------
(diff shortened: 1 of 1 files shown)


raphaelvallat:
Thanks @DavidALloyd, great suggestion. I apologize about the delayed response.

Two things:

1) The `**kwargs` should also be added to the recursive call

https://github.com/raphaelvallat/pingouin/blob/5c5f61a09ee4e49ed81716fa9485527092fcc6b0/pingouin/pairwise.py#L481

and when calculating the interaction:


https://github.com/raphaelvallat/pingouin/blob/5c5f61a09ee4e49ed81716fa9485527092fcc6b0/pingouin/pairwise.py#L540

2) Could you please add unit tests for `mwu`, `wilcoxon`, and `ttest`?

DavidALloyd:
No worries @raphaelvallat, thanks for getting back! I can definitely on adding ```**kwargs``` to the other calls, but I have a bit of a dumb question though about the unit tests. I'm new to setting up tests like this, so my apologies if this is an ignorant question.
>     2. Could you please add unit tests for `mwu`, `wilcoxon`, and `ttest`?

```test_nonparametryic.py``` already seems to contain tests for 'wilcoxon' and 'mwu'

https://github.com/raphaelvallat/pingouin/blob/5c5f61a09ee4e49ed81716fa9485527092fcc6b0/pingouin/tests/test_nonparametric.py#L63

https://github.com/raphaelvallat/pingouin/blob/5c5f61a09ee4e49ed81716fa9485527092fcc6b0/pingouin/tests/test_nonparametric.py#L86

and there's a test for 'ttest' in ```test_parametric.py```

https://github.com/raphaelvallat/pingouin/blob/5c5f61a09ee4e49ed81716fa9485527092fcc6b0/pingouin/tests/test_parametric.py#L33

Was there something specific you wanted me to implement in the testing of ```pairwise_tests``` in ```test_pairwise.py```?

raphaelvallat:
Hi @DavidALloyd!

My message was confusing, sorry about that. We want to add unit tests to [test_pairwise_tests](https://github.com/raphaelvallat/pingouin/blob/5c5f61a09ee4e49ed81716fa9485527092fcc6b0/pingouin/tests/test_pairwise.py#L18) to ensure that the `**kwargs` work as expected. One simple way that doesn't require any ground-truth values from an external software (e..g JASP, JAMOVI, R, Matlab, etc) is to simply ensure that the output of pingouin.pairwise_tests is not the same when using a specific kwarg compared to the default. This needs to be tested for between/within and mixed design (with interaction). You can also just compare the output of `pingouin.pairwise_tests` (with only 2 groups) to the lower-level `mwu`, `wilcoxon`, and `ttest` functions when using a specific kwarg.

Does that make more sense?

Thanks!

raphaelvallat:
Hi,

This PR has been inactive for several years so I'll go ahead and close it. Feel free to re-open.

Raphael

PR #363, opened 2023-06-01
fix groupby.mean for only numeric values

jajcayn:
There was a change in the default behaviour for `groupby(...).mean()`, now the default is `numeric_only=False`


raphaelvallat:
Awesome thanks for catching this! A few comments:

1. We need to bump the pandas dependency to 1.1, since `numeric_only` is not present in 1.0
2. I think there are many other `groupby().mean()` in the Pingouin code. To be safe I think we need to edit all of them.
3. Unit tests are failing for unrelated reasons. I'll fix later.

I'll work on 1 and 2 now. Thanks

Issue #369, opened 2023-07-26
Python3.12: Two tests are failing with AssertionError


penguinpee:
Python 3.12 was unleashed recently in Fedora. Building pingouin against Python 3.12 results in two tests failing:

```python
=================================== FAILURES ===================================
______________________ TestCorrelation.test_partial_corr _______________________
self = <pingouin.tests.test_correlation.TestCorrelation testMethod=test_partial_corr>
    def test_partial_corr(self):
        """Test function partial_corr.
    
        Compare with the R package ppcor (which is also used by JASP).
        """
        df = read_dataset("partial_corr")
        #######################################################################
        # PARTIAL CORRELATION
        #######################################################################
        # With one covariate
>       pc = partial_corr(data=df, x="x", y="y", covar="cv1")
df         =            x         y       cv1       cv2       cv3
0   5.230740  7.141488  1.825430 -1.085631 -0.255619
1   6.013931...681  1.754886
28  3.697089  6.129922  3.138867 -0.140069  1.495644
29  5.466289  7.747689  1.053789 -0.861755  1.069393
self       = <pingouin.tests.test_correlation.TestCorrelation testMethod=test_partial_corr>
pingouin/tests/test_correlation.py:137: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
data =            x         y       cv1       cv2       cv3
0   5.230740  7.141488  1.825430 -1.085631 -0.255619
1   6.013931...681  1.754886
28  3.697089  6.129922  3.138867 -0.140069  1.495644
29  5.466289  7.747689  1.053789 -0.861755  1.069393
x = 'x', y = 'y', covar = 'cv1', x_covar = None, y_covar = None
alternative = 'two-sided', method = 'pearson'
    @pf.register_dataframe_method
    def partial_corr(
        data=None,
        x=None,
        y=None,
        covar=None,
        x_covar=None,
        y_covar=None,
        alternative="two-sided",
        method="pearson",
    ):
        """Partial and semi-partial correlation.
    
        Parameters
        ----------
        data : :py:class:`pandas.DataFrame`
            Pandas Dataframe. Note that this function can also directly be used
            as a :py:class:`pandas.DataFrame` method, in which case this argument
            is no longer needed.
        x, y : string
            x and y. Must be names of columns in ``data``.
        covar : string or list
[...]

PR #370, opened 2023-07-29
Fix in flatten_list for Python 3.12

raphaelvallat:
Closes https://github.com/raphaelvallat/pingouin/issues/369


penguinpee:
I applied the patch and the tests are all passing now. Thanks!

PR #372, opened 2023-08-14
Support for exact tests

ytzfhqs:
Support for exact tests, including fisher exact, barnard exact and boschloo exact

### DIFF
--- a/pingouin/contingency.py
+++ b/pingouin/contingency.py
@@ -5,11 +5,11 @@ import pandas as pd
 
 from scipy.stats.contingency import expected_freq
 from scipy.stats import power_divergence, binom, chi2 as sp_chi2
-
+from scipy.stats import barnard_exact, boschloo_exact, fisher_exact
 from pingouin import power_chi2, _postprocess_dataframe
 
 
-__all__ = ["chi2_independence", "chi2_mcnemar", "dichotomous_crosstab"]
+__all__ = ["chi2_independence", "chi2_mcnemar", "chi2_exact","dichotomous_crosstab"]
 
 
 ###############################################################################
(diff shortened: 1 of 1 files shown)


raphaelvallat:
Hi @ytzfhqs

Thanks for opening the PR! Would you be able to provide some details/justification for adding this to Pingouin? I'm not in favor of adding them since the lower-level scipy functions are already pretty straightforward to use and well documented.

ytzfhqs:
Here is my attempt at translating your message into English:

I'm very happy to receive your reply. I wanted to add exact tests to pingouin for two main reasons: 

1. For situations with expected frequencies less than 5, scipy provides 3 methods - [`fisher_exact`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.fisher_exact.html#scipy.stats.fisher_exact), [`barnard_exact`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.barnard_exact.html#scipy.stats.barnard_exact), and [`boschloo_exact`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.boschloo_exact.html#scipy.stats.boschloo_exact). I wanted to integrate them together as a collection of exact tests. I mentioned this in my previous message, if you remember.

2. I feel like switching back and forth between pingouin and scipy while doing hypothesis testing gives a disjointed feeling. I need to look up documentation in two different places, and the final result formats are also different. Of course this is just my personal opinion, and may not align with pingouin's development principles.

Additionally, I have another question I wanted to ask. scipy's fisher_exact can only be used for 2x2 contingency tables. If my table looks like this:

|      | A    | B    | C    |
| ---- | ---- | ---- | ---- |
| YES  | 11   | 6    | 17   |
| NO   | 13   | 2    | 20   |

How can I use fisher_exact to test if A, B, C have significant effects on the target variable? Could we provide a solution for this in pingouin? I noticed pingouin's chi2_independence uses Yates correction when expected frequencies are <5 - could we prompt the user to use exact tests instead, and provide API interfaces?

Looking forward to your reply!

arthurpaulino:
Oh it's been so long! Unfortunately, at this point, I have less context than @ytzfhqs ):

pingouin/contingency.py
```diff
@@ -345,6 +345,188 @@ def chi2_mcnemar(data, x, y, correction=True):
...
+            odds, pval = (res.statistic, res.pvalue)
+            stats.append(
+            {
+                "alternative": alternative,
+                "odds ratio": odds,
+                "pval": pval,
+            }
+        )
+            
+        stats = pd.DataFrame(stats)[["alternative", "odds ratio", "pval"]]
```
qbarthelemy, line 525: You can easily reduce code redundancy:
```suggestion
    if method == "fisher":
        func = fisher_exact
    elif method == "barnard":
        func = barnard_exact
    elif method == "boschloo":
        func = boschloo_exact
    else:
        raise ValueError(f"Unsupported method {method}")

    for alternative in alternatives:
        res = func(table=observed, alternative=alternative, **kwargs)
        stats.append(
            {
                "alternative": alternative,
                "odds ratio": res.statistic,
                "pval": res.pvalue,
            }
        )

    stats = pd.DataFrame(stats)[["alternative", "odds ratio", "pval"]]
```
A suggestion is to transform `method` argument from a string into any function with a defined signature.

ytzfhqs:
@qbarthelemy  That's a great idea, thank you very much for the suggestion, I will optimise my code!

pingouin/contingency.py
```diff
@@ -345,6 +345,168 @@ def chi2_mcnemar(data, x, y, correction=True):
...
+    observed. There are other possible choices of statistic and two-sided p-value 
+    definition associated with Fisher’s exact test;*
+    
+    *Barnard’s test is an exact test used in the analysis of contingency tables. 
+    It examines the association of two categorical variables, and is a more powerful
+    alternative than Fisher’s exact test for 2x2 contingency tables.*
+    
+    *Boschloo’s test is an exact test used in the analysis of contingency tables.
+    It examines the association of two categorical variables, and is a uniformly 
+    more powerful alternative to Fisher’s exact test for 2x2 contingency tables.*
```
raphaelvallat, line 392: Add:

```
See also
--------
scipy.stats.fisher_exact, scipy.stats.barnard_exact, scipy.stats.boschloo_exact

```

raphaelvallat (changes requested):
Hi @ytzfhqs,

We should be good to merge once you've addressed the two minor comments below.

Thanks,
Raphael

pingouin/contingency.py
```diff
@@ -345,6 +345,168 @@ def chi2_mcnemar(data, x, y, correction=True):
...
+
+    Parameters
+    ----------
+    data : :py:class:`pandas.DataFrame`
+        The dataframe containing the ocurrences for the test.
+    x, y : string
+        The variables names for the exact test. Must be names of columns
+        in ``data``.
+    method : string
+        Methods of exact test. Options include``fisher``,``barnard``,``boschloo``.
```
qbarthelemy, line 360:
```suggestion
    method : {"fisher", "barnard", "boschloo"}, default="fisher"
        Method of exact test.
```

PR #380, opened 2023-09-27
Function test fails for np.mean

### LINKED ISSUE #379: Function test fails for np.mean
Issue seems to affect only Python 3.11

Calling the `compute_bootci` with `np.mean` as the function
```
ci, mean_results = pg.compute_bootci(x, func=np.mean, n_boot=100, return_dist=True, seed=1)
```
fails the function test.
```
AssertionError: func must be a function (e.g. np.mean, custom function) or a string (e.g. 'pearson'). ....
```
It used to work but fails now due to a change in numpy. Here are two relevant `numpy` issues
- https://github.com/numpy/numpy/issues/24019 
- https://github.com/numpy/numpy/issues/23307

The suggested solution is to replace `isfunction` with `isroutine`. 
```
from inspect import isroutine, isfunction
print(isroutine(np.mean))
print(isfunction(np.mean))
```
returns 
```
True
False
```

gedeck:
Closes #379 

Without the change, the 3.11 fails as described in issue #379 . Remaining failures are not introduced by this PR.

### DIFF
--- a/.github/workflows/python_tests.yml
+++ b/.github/workflows/python_tests.yml
@@ -12,7 +12,7 @@ jobs:
       fail-fast: false
       matrix:
         platform: [ubuntu-latest, macos-latest, windows-latest]
-        python-version: [3.7, 3.8, 3.9]
+        python-version: [3.7, 3.8, 3.9, 3.10]
 
     runs-on: ${{ matrix.platform }}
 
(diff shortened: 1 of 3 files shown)


raphaelvallat:
Approved, thank you!

raphaelvallat (approved):

PR #382, opened 2023-10-08
`corr()`: fix `CI95%` column name in returned dataframe

kraktus:
Thanks for the lib, small typo I found when reading the docs.


raphaelvallat (approved):
Approved, thanks!

PR #406, opened 2024-02-16
Switch to modern python packaging

getzze:
Using `setup.py` and  `setup.cfg` is deprecated in favor of a single `pyproject.toml` file:
https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html

This PR removes these two files and the `requirements*.txt` files and puts all the relevant information in `pyproject.toml`.
Also the `pingouin` subfolder is removed, the tests are moved to the `tests` folder and the source to the `src/pingouin` folder.

The version number is uniquely defined in the `__init__.py` source file.


README.rst
```diff
@@ -114,6 +114,18 @@ New releases are frequent so always make sure that you have the latest version:
...
+Development
+-----------
+
+To build and install from source, clone this repository or download the source archive and decompress the files
+
+.. code-block:: shell
+
+  cd pingouin
+  python -m build  # optional, build a wheel
+  pytest           # test the package
```
raphaelvallat, line 126: can we run pytest before doing the `pip install .`?

README.rst
```diff
@@ -114,6 +114,18 @@ New releases are frequent so always make sure that you have the latest version:
...
+  pip install .    # install the package
```
raphaelvallat, line 127: Can you add one line to show how install in editable mode (`pip install --editable .`)

pyproject.toml
```diff
@@ -1,4 +1,116 @@
...
+Homepage = "https://pingouin-stats.org/index.html"
+Downloads = "https://github.com/raphaelvallat/pingouin/"
+
+[tool.setuptools]
+py-modules = ["pingouin"]
+include-package-data = true
+
+[tool.setuptools.package-data]
+pingouin = [
+    "datasets/*.csv",
```
raphaelvallat, line 60: New path is `src/pingouin/datasets`

Also can you please update the datasets paths in `MANIFEST.in`?

raphaelvallat (changes requested):
This is absolutely fantastic @getzze! Thank you for implementing this. I'll definitely re-use this as a template in some of my other projects. One request to update the datasets path and then we should be good to merge.

README.rst
getzze: good point, I'll swap the lines

README.rst
getzze: :+1:

raphaelvallat (approved):
Approved! Thanks again, I'll run some final checks and merge later this week.

raphaelvallat:
@getzze could you please rebase to the latest master, which includes a fix for the unit tests: https://github.com/raphaelvallat/pingouin/pull/403? Thanks

getzze:
I just rebased.

I saw that you don't run the tests on python 3.8 anymore, do you want me to bump the minimal python version in `pyproject.toml` also?

raphaelvallat:
Thank you!

> I saw that you don't run the tests on python 3.8 anymore, do you want me to bump the minimal python version in pyproject.toml also?

I've disabled them because it led to a small rounding error for one particular unit test (logistic regression), most likely related to sklearn version differences. I didn't have time to investigate more, but all the other tests are working fine so I don't think there's a need to bump the minimal version to 3.8.

hoechenberger:
Just FYI with this change, editable installations won't work the way you're used to. We experienced this issue with various MNE-Python-related projects.

The solution was to switch from the `setuptools` to the `hatchling` build backend. It's an easy thing to do and actually a more modern solution. I can help you do the switch either in this PR or in a followup.

See https://github.com/mne-tools/mne-python/issues/12169 for details

raphaelvallat:
@hoechenberger thank you for the heads-up! Very good to know. Let me merge this PR now and feel free to create a new one to use the `hatchling` backend.

Issue #279, opened 2022-06-18
Roadmap for release 0.6.0


raphaelvallat:
The following **issues** should be addressed:

- [x] #308 ❗ 
- [x] #444
- [x] #153
- [x] #253
- [ ] #208
- [x] #225
- [x] #118

The following **PR** should be merged:

- [ ] #226 
- [x] #291
- [x] #443 

The following **dependencies** should be updated:

- [x] Scikit-learn <1.1.2 (revert temporary fix in #278)

The following **deprecation** should be made:

Federico2111:
Seems like a good new release. Is there any time line? When will these upgrades be made?

raphaelvallat:
Hi @Federico2111,

I have had no time to work on Pingouin in the last few months, and I anticipate this will continue at least until Q1 2023. I therefore cannot promise a timeline for this new release.

musicinmybrain:
> The following **dependencies** should be updated:
> 
>     * [ ]  Scikit-learn <1.1.2 (revert temporary fix in [Use scikit-learn<1.1.0 #278](https://github.com/raphaelvallat/pingouin/pull/278))

This can be checked off, since https://github.com/raphaelvallat/pingouin/pull/300 is merged and released in 0.5.3.

Federico2111:
Any update about the progress of this new release? Any help needed?

raphaelvallat:
Thanks @Federico2111 — unfortunately at the moment I only have time for urgent bug fixes, not for implementing new features. So it's unlikely that version 0.6.0 is coming anytime soon; however I'd like to release version 0.5.4 with some minor improvements and bugfixes (e.g. CI tests) in the coming weeks.

Federico2111:
Thanks @raphaelvallat. 
By CI tests, you mean also the parametric Confidence Interval around eta squared that I had suggested? 
https://github.com/raphaelvallat/pingouin/issues/253

raphaelvallat:
No sorry I meant the unit tests (aka continuous integration or CI), see GitHub Actions: https://github.com/raphaelvallat/pingouin/actions

Federico2111:
Any update? :-)

raphaelvallat:
I recently released a minor version that included bugfixes and some improvements: https://github.com/raphaelvallat/pingouin/releases/tag/v0.5.4

I cannot say when the next release will be, but likely not in the coming weeks. Feel free to submit PRs for what you consider to be the most important/urgent tasks. Thanks

[...]

Issue #208, opened 2021-11-02
Naming of return-values for T-tests etc


thomas-haslwanter:
some of the names of the returned dataframe for the function `result = pg.ttest(data)` are unfortunately named, since it does not comply with the naming conventions for variables.
As a result, some parameters can be read out with e.g.
`result.dof`
while those with the non-compliant names can only be accessed with square brackets:
`result['p-value']`

It would be highly desirable to have names that comply with the Python conventions and requirements for variable names.

raphaelvallat:
Thanks @thomas-haslwanter, I agree and I've been wanting to fix this in the next release. We simply need to replace all the "-" in variable names with a "_", e.g. "p-unc" -> "p_unc".

I'll implement that in the next release,

Thanks,
Raphael

thomas-haslwanter:
Don't forget to change the 'CI95%' to 'CI95', since the "%"-sign also causes problems.

thomas-haslwanter:
And one more thing here: parameters that return a single float should not be returned as a pandas Series object, but simply as a float.
For example, the p-value of the test
`result = pg.ttest(before, after)`
currently has to be retrieved as
`result['p-value']['T-test']`
This should be simplified to
`result.pval`

raphaelvallat:
Hi @thomas-haslwanter,

This would mean returning the output of most Pingouin functions as a pandas.Series instead of a pandas.DataFrame. While it would be indeed simpler to access the value, I think that the Series output in Jupyter notebook is less easy-to-read than a traditional DataFrame. This is quite a big conceptual modification, so we should discuss that in a separate issue and maybe do a poll.

Thanks,
Raphael

thomas-haslwanter:
When thinking it though I actually agree with you: using a pd.DataFrame for the result really makes the results MUCH clearer, and should therefore be kept. Either way, thank you for the quick reply!

raphaelvallat:
Edit: this will not be included in the next release of Pingouin (v0.5.1) which is a minor release, but it should be included in the next major release (0.6.0).

PR #443, opened 2024-10-07
Pandas-friendly column names

remrama:
This PR addresses #208 (part of the `v0.6.0` Roadmap layed out in #279). It removes characters that restrict column access to the bracket format (`df["p_val"]`) rather than dot(?) method (`df.p_val`). It includes the following column-name changes:

1. Replaced dashes with underscores in column names (e.g., `p-val` --> `p_val`)
2. Replaced parentheses with underscores in column names (e.g., `mean(A)` --> `mean_A`)
3. Removed percent signs and brackets in column names (e.g., `CI95%` --> `CI95`, `CI[97.5%]` --> `CI97.5`)

Here is a specific list of most (if not all) of the changes:

- `p-adjust` --> `p_adjust`
- `p-approx` --> `p_approx`
- `p-corr` --> `p_corr`
- `p-exact` --> `p_exact`
- `p-mid` --> `p_mid`
- `p-spher` --> `p_sphere`
- `p-tukey` --> `p_tukey`
- `p-unc` --> `p_unc`
- `p-val` --> `p_val`
- `U-val` --> `U_val`
- `W-spher` --> `W_spher`
- `W-val` --> `W_val`
- `p-GG-corr` --> `p_GG_corr`
- `cohen-d` --> `cohen_d`
- `eta-square` --> `eta_square`
- `odds-ratio` --> `odds_ratio`
- `CI95%` --> `CI95`
- `CI90%` --> `CI90`
- `CI[2.5%]` --> `CI2.5`
- `CI[97.5%]` --> `CI97.5`
- `mean(A)` --> `mean_A`
- `mean(B)` --> `mean_B`
- `std(A)` --> `std_A`
- `std(B)` --> `std_B`

### Future column-name considerations

In doing this, I noted some other column-name considerations for future PRs that feel less immediate:

1. Standardizing the use of "val". Modules differentially name p-value columns as `p`, `pval`, or `p_val`. This should be standardized across modules.
2. Related to (1), the norm for p-values might correspond with other "values", like the current column names `T`, `F`, `W`, `r`, etc. I think optimally this would either be `pval` and `Wval`/`wval` or `p` and `W`/`w`. Perhaps an argument for keeping "val" is that `.T` blocks the pandas transpose method.
3. Standardize some effect size namings across modules. Cohen's d is differentially referred to as `cohen` or `cohen_d`, and eta-squared is differentially referred to as `eta_square`, `eta-squared`, or `n2`.
4. Standardize the degrees of freedom column names. These are differentially named `DF`, `dof`, `ddof`.
5. Standardize abbreviation capitalizations. Bayes factors are `BF`, confidence intervals are `CI`, Wilcoxon stats are `W`, then there's `dof` and `r`.
6. Standardize Sentence case vs camelCase vs snake_case. Column capitalization, like `Parametric` vs `alternative` are inconsistent. I think snake_case is the pandas norm, but consistency seems most important.
7. I think including the numerical values in confidence interval column names is not necessary. Eg, `CI95` could just be `CI`, and, more importantly, `CI97.5` and `CI2.5` could just be `CI_upper` and `CI_lower`, respectively.

### Notes

* The jupyter notebooks should run fine with the new code, but I did not re-run them, so the output of those tutorials still have the old column names. **Let me know if you want me to rerun them before merging.**
* I updated the docstrings as needed, but did not build them locally to test. **Let me know if you want me to build and inspect before merging.**


raphaelvallat:
@remrama this is absolutely **fantastic**. Thank you for doing such a thorough investigation. My thoughts on what you described in the "Future column-name considerations":

1. Overall, I think we should merge the current PR and then implement most, if not all of the changes that you describe, and then release a new 0.6.0 version. The other items in the roadmap feel less important and could be included as part of the 6.x minor version updates.
2. Agreed for:
    - using `*val` throughout
    - use `cohen_d` and `eta_squared` (and replace `np2` with `partial_eta_squared`?)
    - use `ddof` to be consistent with pandas and scipy
    - use snake_case throughout
    - use `ci`, `ci_upper`, `ci_lower` throughout
    - perhaps even worth adding a new documentation page with Pingouin's standard naming conventions...?

Laslty, yes to both your questions in the "Notes" section. :)

remrama:
Great, thanks @raphaelvallat. I think the current PR is now ready for a quick review. I updated the notebooks and tested the docs locally (passed build and visual inspection). I'm a bit confused as to why the Python tests aren't running after a PR (Lint is, but not others). I ran tests locally and they passed, but I can't speak for alternate systems and Python versions.

Happy to push for another PR that adds to the column-name conventions after this one :)

raphaelvallat:
**Fix the Github Action CI:**

Replace 
```yaml
name: Python tests

on:
  push:
    branches: [master, develop]
  pull_request:
    branches: [master, develop]
```

with:

```yaml
name: Python tests

on:
  push:
    branches: [main]
  pull_request:
    branches: [main]
```

raphaelvallat:
I'll wait for the unit tests and then do a final review and approval. Hopefully we don't get failing unit tests from previous PRs 🤦 Feel free to re-enable the CI in a separate PR, which we'll merge before this one

raphaelvallat (approved):
Thank you so much @remrama , approved! 

> I had a last-minute thought about this PR: It really doesn't impact at all how the user interfaces with Pingouin (only the output). But there is one exception with the compute_effsize function, where a few parameters won't work anymore. Passing in eta-square and odds-ratio now will not work, user needs to pass eta_square or odds_ratio. Maybe just leave it, but I was thinking we could add a simple "-".replace("_") and a FutureWarning. The only reason these were changed is because they end up as column names in some functions, so the other option would be to just name the columns something like effsize and not carry the actual stat name over.

I think let's not worry about the FutureWarning and `replace`. We should just leverage this major update to do a clean slate.

remrama:
@raphaelvallat this PR (and #446) have been approved for a bit, I think they can be merged? Lmk if I am missing something I need to do. Otherwise I just want to make sure they don't fall behind and end up with conflicts.

raphaelvallat:
Thanks for the reminder! Merging now!

PR #451, opened 2024-12-04
Clean dependencies; add extras feature

getzze:
Summary:
* reorder `pyproject.toml` entries to make it easier to find the information.
* add `python 3.12` classifier
* remove `codecov`, `openpyxl` ~and `mpmath`~ test dependencies as they are not used in tests.
* add `extras` features to optionally install the `mpmath` dependency with `pip install pingouin[extras]`.

On a side note, `mpmath` is an optional dependency as it is checked if it's installed in `utils.py`.
It's also checked if `scikit-learn` and `statsmodels` are installed in `utils.py`, whereas they are part of the hard dependencies. Meaning they _should_ be installed, at least if installing with `pip`. These two dependencies could be moved to `extras` dependencies or the checks in `utils.py` could be removed.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -48,6 +48,7 @@ test = [
     "pytest-cov",
     # Ensure coverage is new enough for `source_pkgs`.
     "coverage[toml]>=5.3",
+    "mpmath",
 ]
 docs = [
     "sphinx>7.0.0",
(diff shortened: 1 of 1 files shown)


raphaelvallat (approved):
Thank you — this is great. I prefer to keep statsmodels and scikit-learn as hard dependencies. I'll merge the PR now.

PR #457, opened 2025-01-20
Update RBC calculation for Wilcoxon signed-rank test to be dependent on the alternative

### LINKED ISSUE #45: Add support for three-way ANOVA
Using statsmodels:
- https://www.statsmodels.org/stable/anova.html
- https://www.statsmodels.org/stable/generated/statsmodels.stats.anova.anova_lm.html

### LINKED ISSUE #456: Matched-pairs rank biserial correlation for the wilcoxon signed-rank test
(Summarizing the discussion from https://github.com/raphaelvallat/pingouin/discussions/455)

From my understanding, the matched-pairs rank biserial correlation according to Kerby 2014 is calculated as the rank sum difference between favorable and unfavorable evidence for the Wilcoxon signed-rank test (see for example Table 2 on p. 6). This means wether or not a difference in value is considered favorable or unfavorable should depend on the hypothesis.

In Pingouin, differences larger than zero are always considered favorable evidence.

Kerby (2014) states that:

> In words, the formula states that the nonparametric correlation equals the simple difference between the proportion of favorable and unfavorable evidence; in the case of the Mann-Whitney U test, the evidence consists of pairs.
An advantage of the simple difference formula is that it expresses one effect size (the rank-biserial correlation) in terms of another easily understood measure of effect size (the common language effect size). The other formulas for the rank-biserial express it in terms that are less easily interpreted. **A second advantage of the simple difference formula is that it gives meaning to the direction of the sign. A positive correlation means that the data were in the predicted direction; a negative correlation, that they were against the predicted direction.**

rhazn:
Closes #456 

I had a look at providing a PR, but I am unsure if I broke more than I assumed or I am misunderstanding something. If I rerun the examples from the documentation, I get some different p-values (I updated the docs with my values in this PR) and I am seeing a test failure with just slightly mismatched values in `tests/test_pairwise.py:435`:

```E       Mismatched elements: 3 / 3 (100%)
E       Max absolute difference among violations: 0.021
E       Max relative difference among violations: 0.04347826
E        ACTUAL: array([0.912, 0.96 , 0.462])
E        DESIRED: array([0.91 , 0.951, 0.483])```

I will be able to spend some more time on this in a few days, but until then I was wondering if this might be related to a setup issue from my side you can immediately spot and help me remove?

### DIFF
--- a/src/pingouin/nonparametric.py
+++ b/src/pingouin/nonparametric.py
@@ -410,37 +410,37 @@ def wilcoxon(x, y=None, alternative="two-sided", **kwargs):
     >>> y = np.array([38, 37, 33, 29, 14, 12, 20, 22, 17, 25, 26, 16])
     >>> pg.wilcoxon(x, y, alternative='two-sided')
               W_val alternative     p_val       RBC      CLES
-    Wilcoxon   20.5   two-sided  0.285765 -0.378788  0.395833
+    Wilcoxon   20.5   two-sided  0.288086 -0.378788  0.395833
 
     Same but using pre-computed differences. However, the CLES effect size
     cannot be computed as it requires the raw data.
 
     >>> pg.wilcoxon(x - y)
               W_val alternative     p_val       RBC  CLES
-    Wilcoxon   20.5   two-sided  0.285765 -0.378788   NaN
+    Wilcoxon   20.5   two-sided  0.288086 -0.378788   NaN
 
     Compare with SciPy
 
     >>> import scipy
     >>> scipy.stats.wilcoxon(x, y)
-    WilcoxonResult(statistic=20.5, pvalue=0.2661660677806492)
+    WilcoxonResult(statistic=20.5, pvalue=0.2880859375)
 
     The p-value is not exactly similar to Pingouin. This is because Pingouin automatically applies
     a continuity correction. Disabling it gives the same p-value as scipy:
 
     >>> pg.wilcoxon(x, y, alternative='two-sided', correction=False)
               W_val alternative     p_val       RBC      CLES
-    Wilcoxon   20.5   two-sided  0.266166 -0.378788  0.395833
+    Wilcoxon   20.5   two-sided  0.288086 -0.378788  0.395833
 
     One-sided test
 
     >>> pg.wilcoxon(x, y, alternative='greater')
               W_val alternative     p_val       RBC      CLES
-    Wilcoxon   20.5     greater  0.876244 -0.378788  0.395833
+    Wilcoxon   20.5     greater  0.865723 -0.378788  0.395833
 
     >>> pg.wilcoxon(x, y, alternative='less')
               W_val alternative     p_val       RBC      CLES
-    Wilcoxon   20.5        less  0.142883 -0.378788  0.604167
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by remrama** [comment]
Hi @rhazn, are you sure this is to close 45? Maybe that was a typo because 45 deals with ANOVAs and was closed years ago. Please edit the original post to match the relevant Issue if I'm right.

As far as the failing tests, when I make the same single change in L493 as you, I get the same output for all docs/tests as were originally there. Like, it doesn't require the small changes you added. But you passed 3.9 tests (only), so I guess those changes helped with that? I'm not sure fully what the deal is, but just letting you know what I found after a quick check. I'm using Python 3.11 on Windows for what that's worth.

**Comment by rhazn** [comment]
Ah sorry, you are correct. I think I accidentally deleted a number, this PR is in reference to #456.

Sadly, I am not well versed in Python development so I might struggle more with this than average. I am working on OSX, when I run the docs examples with 3.11 in my REPL, I get the following output:

```
Python 3.11.11 (main, Jan 14 2025, 23:36:41) [Clang 19.1.6 ] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import numpy as np
>>> import pingouin as pg
>>> x = np.array([20, 22, 19, 20, 22, 18, 24, 20, 19, 24, 26, 13])
>>> y = np.array([38, 37, 33, 29, 14, 12, 20, 22, 17, 25, 26, 16])
>>> pg.wilcoxon(x, y, alternative='two-sided')
          W_val alternative     p_val       RBC      CLES
Wilcoxon   20.5   two-sided  0.288086 -0.378788  0.395833
>>> pg.wilcoxon(x - y)
          W_val alternative     p_val       RBC  CLES
Wilcoxon   20.5   two-sided  0.288086 -0.378788   NaN
```

Though with 3.9:
```
Python 3.9.21 (main, Jan 14 2025, 23:33:15) 
[Clang 19.1.6 ] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import numpy as np
>>> import pingouin as pg
>>> x = np.array([20, 22, 19, 20, 22, 18, 24, 20, 19, 24, 26, 13])
>>> y = np.array([38, 37, 33, 29, 14, 12, 20, 22, 17, 25, 26, 16])
>>> pg.wilcoxon(x, y, alternative='two-sided')
          W_val alternative     p_val       RBC      CLES
Wilcoxon   20.5   two-sided  0.285765 -0.378788  0.395833
>>> pg.wilcoxon(x - y)
          W_val alternative     p_val       RBC  CLES
Wilcoxon   20.5   two-sided  0.285765 -0.378788   NaN
```

Maybe this is something that is not good to discuss in a PR and should go into a Q&A, but it feels like I am making a mistake with Python venv that is preventing me from providing a good PR 😅 .

remrama:
Hmm, I'm not sure what is going on between the Python versions either... 🤔

raphaelvallat (approved):
Thanks for the PR! I've fixed (= disabled) the unrelated failing tests in https://github.com/raphaelvallat/pingouin/pull/460. I believe they were caused by differences in the p-value calculation in the scipy.stats.wilcoxon function starting from Scipy 1.14+. Merging this PR now.

PR #333, opened 2023-01-05
Adding to description for non parametrics array

turkalpmd:
I got problems and solved it but this time I received Lint error, I will fixing.

PR #464, opened 2025-04-06
#333 Look PR comments (https://github.com/raphaelvallat/pingouin/pull…

turkalpmd:
…/333)

I never thought we would be delayed by almost 5–6 months. I had to adapt to a new country — I’m sorry.

https://github.com/raphaelvallat/pingouin/pull/333#issuecomment-2395557623

### DIFF
--- a/src/pingouin/contingency.py
+++ b/src/pingouin/contingency.py
@@ -9,7 +9,7 @@ from scipy.stats import power_divergence, binom, chi2 as sp_chi2
 from pingouin import power_chi2, _postprocess_dataframe
 
 
-__all__ = ["chi2_independence", "chi2_mcnemar", "dichotomous_crosstab"]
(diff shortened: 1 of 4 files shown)


turkalpmd:
I cant fix this error. I cant understand, because it is already succesfully passed in my local.

raphaelvallat:
@turkalpmd can you please limit the scope of this PR to the improvements of the non-parametric pairwise T-tests, and remove the Ransacking post-hoc analysis (which should be implemented in a separate PR)? Thanks

turkalpmd:
As you suggested, I’ve now created separate PRs for each branch. 
Here is the Ransacking method: https://github.com/raphaelvallat/pingouin/pull/465
I think this is clearly passed to everything. Now we should fix this PR, I can't understand why I couldn't pass this.

PR #465, opened 2025-04-10
Ransacking

turkalpmd:
As you suggested, I’ve now created separate PRs for each branch. This PR is for the Ransacking method.


raphaelvallat:
Hi,

Sorry for the late reply. Pingouin isn’t under active development anymore, so I’m not adding new major features at this point. I’m focusing on bug fixes and dependency updates instead.

I’ll close this PR for now. Thanks for understanding,

Raphael

PR #466, opened 2025-04-29
Implement Mardia–Watson–Wheeler test

GanshengT:
Regarding issue #214, I implemented the Mardia–Watson–Wheeler test in Pingouin, hope that is a useful addition.


GanshengT:
The test did not pass due to calling the wrong function

PR #474, opened 2025-07-12
Fixes #453: _correl_pvalue does not divide by zero if r==1

### LINKED ISSUE #453: divide by zero encountered to calculate p-value when r==1
In correlation.py the function `_correl_pvalue(r, nx, k=0) `does not handle the case when r=1.0.

This results in: 
```
lib/python3.12/site-packages/pingouin/effsize.py:152): RuntimeWarning: divide by zero encountered in arctanh
  z = np.arctanh(stat)  # R-to-z transform

lib/python3.12/site-packages/pingouin/correlation.py:68): RuntimeWarning: divide by zero encountered in scalar divide
  tval = r * np.sqrt(dof [/](https://file+.vscode-resource.vscode-cdn.net/) (1 - r**2))
```

AlexanderJCS:
Before this change, `_correl_pvalue` would create a division by zero warning when r==1.

Simply adding the condition near the top of the function:
```py
    if np.isclose(r**2, 1):  # Avoid divide by zero error
        return 0.0  # Since p value approaches 0 as r approaches 1, just return 0
```

Fixes this. This assumes that n >= 2 since p would be undefined if n < 2, but I did not account for this case since it seems that the function already assumes that n >= 2.

Please let me know if you want me to make any other changes.

### DIFF
--- a/src/pingouin/correlation.py
+++ b/src/pingouin/correlation.py
@@ -54,6 +54,9 @@ def _correl_pvalue(r, n, k=0, alternative="two-sided"):
         "less",
     ], "Alternative must be one of 'two-sided' (default), 'greater' or 'less'."
 
+    if np.isclose(r**2, 1):  # Avoid divide by zero error
+        return 0.0  # Since p value approaches 0 as r approaches 1, just return 0
+
     # Method 1: using a student T distribution
     dof = n - k - 2
     tval = r * np.sqrt(dof / (1 - r**2))
--- a/tests/test_correlation.py
+++ b/tests/test_correlation.py
@@ -115,13 +115,13 @@ class TestCorrelation(TestCase):
         stats = corr(x, x)
         assert np.isclose(stats.at["pearson", "r"], 1)
         assert np.isclose(stats.at["pearson", "power"], 1)
-        
+
         # Perfect correlation with percbend method
-        #https://github.com/raphaelvallat/pingouin/issues/453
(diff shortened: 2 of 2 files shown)


src/pingouin/correlation.py
```diff
@@ -54,6 +54,9 @@ def _correl_pvalue(r, n, k=0, alternative="two-sided"):
         "less",
     ], "Alternative must be one of 'two-sided' (default), 'greater' or 'less'."
 
+    if np.isclose(r**2, 1):  # Avoid divide by zero error
+        return 0.0  # Since p value approaches 0 as r approaches 1, just return 0
+
```
raphaelvallat, line 59: Can you please add a test in https://github.com/raphaelvallat/pingouin/blob/main/tests/test_correlation.py `test_corr` function:

```python
stats = corr(x, x, method="percbend")  # calls _correl_pvalue
assert np.isclose(stats.at["percbend", "r"], 1)
assert np.isclose(stats.at["percbend", "p_val"], 0)
```

raphaelvallat (changes requested):
Hi @AlexanderJCS — apologies for the long delay. One request to update the tests to make sure this line is covered by the tests. Thanks

src/pingouin/correlation.py
AlexanderJCS: Hi Raphael, I added these lines in the `test_corr(self)` function under the perfect correlation test for the pearson method.

PR #476, opened 2025-08-10
Fix rank biserial correlation sign for Mann-Whitney U test

thimo72:
Changed RBC calculation from
    rbc = 1 - (2 * uval) / diff.size
to
    rbc = 2 * cles - 1

This makes RBC directional and consistent with CLES, so that RBC > 0 when x > y.


raphaelvallat:
Hi,

Is this linked to a particular issue? If not, can you please first open an issue to describe the problem, with minimal code to reproduce the issue? 

Also, for this PR, please add unit tests to validate the new behavior.

Thanks
Raphael

PR #480, opened 2025-12-31
Upgrade supported python versions and support pandas 3.x

yann1cks:
Hello @raphaelvallat,

this PR adds "official" support for newer Python versions up to 3.14 and drops support for EOL versions. 

I removed one unit test for `**kwargs` in `logistic_regression`, because `multi_class` is not supported in newer scikit-learn versions. There is another test for kwargs anyway. 

Moreover I changed the max supported scikit-learn version to <1.8, because of changes to `MinCovDet` in https://github.com/scikit-learn/scikit-learn/issues/23162. The results of `skipped` correlations do not match the Matlab implementation (at least based on the unit test). Currently the function returns incorrect results for all users that use scitkit-learn 1.8, so a quick update is needed. Alternatively we could add a check in the `skipped` function and raise an Exception if the sklearn-version is 1.8 or above.

Unfortunately I do not have access to Matlab to compare the results and do not have the time to look into the `skipped` correlation function in detail. It should be possible to fix it for newer sklearn versions and use different code paths. 

Based on the my problems I suggest to add unit tests against mayor and minor versions (including release candidates) of essential dependencies (pandas, scikit-learn, statsmodels, numpy, scipy). The test suite is very fast so and additional weekly run of wide test suite might help to catch similar problems early.

### DIFF
--- a/src/pingouin/correlation.py
+++ b/src/pingouin/correlation.py
@@ -1097,7 +1097,6 @@ def rcorr(
         else:
             # Method = 'spearman'
             mat_upper = self.corr(method=lambda x, y: spearmanr(x, y)[1], numeric_only=True)
-
         if padjust is not None:
             mask = np.triu(np.ones(mat.shape, dtype=bool), k=1)
             pvals = np.where(mask, mat_upper.to_numpy(), 0)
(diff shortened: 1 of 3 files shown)


raphaelvallat:
Amazing, thank you @yann1cks ! Let's not worry about the code coverage failure, I'll go ahead and merge the PR. Thanks

PR #489, opened 2026-02-16
Update build infra; better testing with minimal deps; ruff format

yann1cks:
Hello @raphaelvallat,

this PR includes the updates I mentioned in #480 and some additional changes:
- Uses dependencies groups and not optional dependencies for dev dependencies. See [Handbook ](https://pydevtools.com/handbook/explanation/what-are-optional-dependencies-and-dependency-groups/) for more info. Perhaps statsmodels and scikitlearn could become optional dependencies as well?
- Updates the ruff config to sort imports and foramt doc strings as well. I did not enable stricter rules, that would require code changes, e.g. replacing pandas `.at`
- Improve docs to directly link to source code on Github (function made with Claude).
- Separate Github Actions in multiple parts and use `uv` to speed it up. I removed the codecov token, because it is publicly available. I think you should rewoke this token.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -29,13 +29,13 @@ dynamic = ["version"]
 requires-python = ">=3.10"
 dependencies = [
     "matplotlib",
-    "numpy",
-    "pandas>=1.5",
+    "numpy>=1.22.4",
+    "pandas>=2.1.1",
     "pandas_flavor",
-    "scikit-learn>=1.2",
-    "scipy",
+    "scikit-learn>=1.2.2",
+    "scipy>=1.8.0",
     "seaborn",
-    "statsmodels",
+    "statsmodels>=0.14.1",
     "tabulate",
 ]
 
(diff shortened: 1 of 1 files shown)


yann1cks:
There are two different pytest setups. 

One is called core and similar to the old one, but I removed some python versions. It tests different os and python versions with the newest dependency versions. 

The other one defines hard coded dependency versions for different python versions and for different os versions as well. For python 3.10 I tried to select the oldest dependencies possible/sensible. This needs more maintenance. But it already shows that pingouin is not compatible with old pandas and numpy versions. There is also a weird dependency resolution error with statsmodels. 

This leads to quite a lot of combinations. What do you think about that? Is it too much? Perhaps we could remove macOS or reduce the python versions for the core tests. 

(Moreover there is pre release setup that only runs on schedule or manually.)

raphaelvallat:
@yann1cks please see my edits. All tests are passing ✅ Note that I had to bump pandas to 2.1.1 and statsmodels to 0.14.1 to avoid a bug on windows.

pyproject.toml
yann1cks: I think the minimal package versions from the tests should be used for the dependencies  as well.

PR #515, opened 2026-06-28
Add brunner_munzel(): nonparametric test for stochastic equality (Brunner & Munzel 2000)

hass-nation:
## Summary

Adds `brunner_munzel(x, y)` — the Brunner-Munzel test, a nonparametric test for stochastic equality of two independent groups. It is more robust than the Wilcoxon rank-sum / Mann-Whitney U test under heteroscedasticity (unequal variances), which is its main practical advantage.

## Algorithm

Uses a Welch-Satterthwaite degrees-of-freedom approximation and a t-distribution reference. The test statistic is based on within-group and between-group rank differences. p-values are verified to match `scipy.stats.brunnermunzel`.

## Interface

```python
from pingouin import brunner_munzel

brunner_munzel(x, y, alternative='two-sided')
# Returns DataFrame with columns: W-val, dof, alternative, p-val, RBC, CLES
# CLES = P(X > Y), the common language effect size
# RBC  = rank-biserial correlation
```

## New content

- `brunner_munzel()` function added to `pingouin/nonparametric.py` and exported from `pingouin/__init__.py`
- 11 tests added to `tests/test_nonparametric.py` covering: basic correctness vs scipy, identical groups, clear-difference groups, one-sided alternatives, effect size bounds, all three `nan_policy` modes, output column names

## Reference

Brunner, E. & Munzel, U. (2000). The nonparametric Behrens-Fisher problem: asymptotic theory and a small-sample approximation. *Biometrical Journal*, 42(1), 17–25.

### DIFF
--- a/src/pingouin/nonparametric.py
+++ b/src/pingouin/nonparametric.py
@@ -15,6 +15,7 @@ __all__ = [
     "friedman",
     "cochran",
     "harrelldavis",
+    "brunner_munzel",
 ]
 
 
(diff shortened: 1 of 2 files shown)


raphaelvallat:
Hi,

Thanks for opening the PR. Brunner Munzel is already supported by [scipy.stats](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.brunnermunzel.html), so I don't see a strong reason to add it to Pingouin (I'm not sure what the value add would be), unless others feel strongly about it.

PR #520, opened 2026-09-08
Fix scale-dependent standard errors and p-values in linear regression

dnncha:
`linear_regression()` computes coefficients with `lstsq(Xw, yw)`, but computes their covariance with `pinv(Xw.T @ Xw)`. Forming the normal matrix squares the condition number. Its pseudoinverse can discard a direction that the coefficient fit retains, giving an extremely small standard error and a reported p-value of zero for an ordinary noise association.

Compute the covariance diagonal from the whitened design's SVD, retaining the rank already used by `lstsq`. This preserves the coefficient fit, residual variance and degrees of freedom, and avoids forming the normal matrix.

Reproduction: generate independent length-100 normal vectors with `np.random.default_rng(42)` and fit y against x. The original slope p-value is 0.349698. Multiplying x by 1e-8 changes it to 0; SciPy `linregress` preserves 0.349698. Across seeds 0–99 with independent normal x and y, original rejection counts at 0.05 are 3/100 before rescaling and 100/100 after; SciPy and the correction both give 3/100 in either representation. These are finite simulation results, not a population error-rate estimate.

A separate full OLS check on all 442 rows and ten predictors of `sklearn.datasets.load_diabetes(scaled=False)` gives HDL (`s3`) p≈0.634723. Rescaling only `s3` by 1e-6 makes the original standard error 0.002463 instead of about 782,464, and p=0. The correction matches statsmodels OLS, including the confidence interval and p≈0.634723 at each tested scale. This is a representation-invariance check on public example data, not a claim about the original study's conclusions or physical units of the distributed values.

Validation on upstream base `35f88a7e90f9e801ad0cf92e25605fea3fb3192c`: 15 added cases cover ordinary/weighted fits, intercept on/off and scales 1, 1e-8 and 1e8 against statsmodels, plus SciPy simple-regression references. Before: 10 fail and five pass. After: the complete regression test file gives **18 passed**, including existing rank-deficient and mediation checks. Ruff check and format checks pass. Released Pingouin 0.6.1 reproduces the defect. The full package suite was not run.

Prepared with AI assistance. No maintainer acceptance or independent scientific review is claimed. The [Cheerful Duck investigation](https://cheerfulduck.com/research/audits/pingouin-regression-unit-scaling) provides the reproduction script, patch, environment, data checksum and before/after execution evidence in a downloadable archive.

### DIFF
--- a/src/pingouin/regression.py
+++ b/src/pingouin/regression.py
@@ -415,8 +415,7 @@ def linear_regression(
     # (machine epsilon) is too strict and lets exactly collinear designs pass
     # as full rank, so we use the same tolerance as numpy.linalg.matrix_rank.
     rcond = max(Xw.shape) * np.finfo(float).eps
-    coef, ss_res, rank, _ = lstsq(Xw, yw, cond=rcond)
-    ss_res = ss_res[0] if ss_res.shape == (1,) else ss_res
+    coef, _, rank, _ = lstsq(Xw, yw, cond=rcond)
     if coef_only:
         return coef
     if rank < Xw.shape[1]:
@@ -435,10 +434,9 @@ def linear_regression(
     # Calculate predicted values and (weighted) residuals
     pred = Xw @ coef
     resid = yw - pred
-    if np.size(ss_res) == 0:
-        # lstsq returns an empty ss_res when the design is rank deficient or
-        # when n <= p, so compute it from the residuals instead.
-        ss_res = (resid**2).sum()
+    # Do not rely on the residues returned by lstsq: depending on the SciPy
+    # version they are empty or NaN for rank-deficient and n <= p designs.
+    ss_res = (resid**2).sum()
 
     # Calculate total (weighted) sums of squares and R^2
     ss_tot = yw @ yw
--- a/src/pingouin/regression.py
+++ b/src/pingouin/regression.py
@@ -449,7 +449,12 @@ def linear_regression(
 
     # Compute mean squared error, variance and SE
     mse = ss_res / df_resid
-    beta_var = mse * (np.linalg.pinv(Xw.T @ Xw).diagonal())
+    # Inverting Xw.T @ Xw squares the condition number and can discard
+    # estimable directions when predictors have different units. Form the
+    # covariance from the design SVD, retaining the rank used by lstsq.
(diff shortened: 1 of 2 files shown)


raphaelvallat:
Thank you for catching this @dnncha ! 

Claude review flagged a few other issues that I have addressed in two commits:

- **Explicit rank tolerance in `lstsq`.** The covariance now relies on the rank reported by `lstsq`, but its default tolerance (`cond=None`, machine epsilon) lets exactly collinear designs, such as an intercept plus a full set of one-hot dummies, pass as full rank on many seeds. No rank-deficiency warning fired and `coef`/`se` came out around 1e13. We now pass `cond = max(n, p) * eps` (the same tolerance `numpy.linalg.matrix_rank` uses), and I added a test on a seed that previously slipped through. It now warns and matches statsmodels.
- **Saturated designs (n == p).** With a square full-rank design, `lstsq` returns an empty `ss_res` and the new covariance line raised a broadcast `ValueError`. This was pre-existing, but on the same line, so I fixed it here: `ss_res` is now recomputed from the residuals whenever `lstsq` returns it empty, and the result is `se=inf` / `pval=NaN`, consistent with the n < p case. Same commit also passes `copy=False` to the `astype` before the SVD (avoids a redundant n x p copy) and drops the seed-dependent `pval > 0.3` assertion, which the exact match against `linregress` above it already implies.

Update: SciPy 1.18 changed `lstsq` to return NaN residues for rank-deficient designs, which broke the 3.12/3.14 CI tests. Fixed in 9ad765b by always computing the residual sum of squares from the residuals.

Please feel free to review my changes, and we'll merge once I have your green light.

raphaelvallat:
See also https://github.com/raphaelvallat/pingouin/issues/522 (will be addressed in future PRs)

dnncha:
Thanks Raphael, I’ve reviewed the changes and I’m happy for you to merge. The explicit rank tolerance and recomputing residuals both make sense.

All 20 regression tests pass locally, plus 20 extra comparisons with statsmodels across weighted and unweighted collinear designs. I’ve also read #522; those follow-ups can stay separate from this fix. Thanks for checking these edge cases.

raphaelvallat (approved):

PR #521, opened 2026-09-08
Fix rcorr multiple-testing family by excluding matrix placeholders

dnncha:
`rcorr(..., padjust=...)` currently passes a square matrix to `multicomp`, filling the diagonal and unused lower triangle with zero p-values. These placeholders enter the test family. For 20 variables, this adds 210 fictitious tests to 190 real unique-pair tests. BH can become severely liberal; Bonferroni/Sidak instead overcorrect. Default `padjust=None` is unaffected.

This change uses NaN for unused cells, which `multicomp` already excludes. It preserves the returned table layout and raw correlations.

Reproduction on released 0.6.1 and upstream 35f88a7e90f9e801ad0cf92e25605fea3fb3192c:

```python
import numpy as np
import pandas as pd
import pingouin as pg
from scipy.stats import pearsonr, false_discovery_control
x = pd.DataFrame(np.random.default_rng(42).normal(size=(100, 20)))
i, j = np.triu_indices(20, 1)
p = [pearsonr(x[a], x[b]).pvalue for a, b in zip(i, j)]
print(false_discovery_control(p).min())  # 0.2545917396820073
print(pg.rcorr(x, padjust='fdr_bh', stars=False, decimals=12))
# Original minimum: 0.002540201942, with four significant pairs (also starred).
# Corrected: matches the unique-pair reference; no significant pairs.
```

Across 100 fixed independent-normal datasets (seeds 0–99, 100 rows × 20 variables), the original produces 525 rejections in 100/100 families versus four rejections in 4/100 families using SciPy BH on the actual pairs. These are finite simulation results, not an exact population FDR or an estimate of affected publications. A controlled feature-correlation analysis of scikit-learn's Wisconsin diagnostic data changes 389 to 388 significant pairs; the diabetes feature table remains 41/45 in both versions. No original-paper conclusion or clinical effect is established.

Validation: 25 new parameterized regression checks cover Pearson/Spearman, all five correction methods, unadjusted controls, missing observations and displayed stars. Before: 15 fail, 10 pass. The correction makes all 25 pass. Replays of 100 null tables and both public datasets agree with independent SciPy BH after correction. Maintainer review and a released fix remain pending.

The combined new tests and existing `test_pandas.py` / `test_correlation.py` suite passes all 30 tests. An initial broader run lacked optional `mpmath`; after installing it, the suite passes. Ruff lint and formatting checks pass. This is focused validation, not a full-package test run. Prepared with AI assistance.

Public report and source-hashed reproduction archive: https://cheerfulduck.com/research/audits/pingouin-correlation-test-family . The report includes all 100 null cases, the changed breast-cancer feature decision, the unchanged diabetes control, complete source before/after, tests and environment records.


raphaelvallat:
Thanks for the report and the fix @dnncha !

I pushed two follow-up commits to this branch to address related issues that came up in AI review:

f9d1e0b – Share upper-triangle p-value adjustment between rcorr and ptests
- Added a private helper `_multicomp_triu` in `multicomp.py` that adjusts only the n(n-1)/2 unique pairs of a square p-value matrix. This makes the test family explicit at the call site rather than relying on NaN placeholders being ignored inside multicomp.
- rcorr now uses the helper (same results as the NaN-padded version in your commit).
- ptests had the same bug from #480 but with a twist: it hardcoded method="bonf" and ignored the user's padjust entirely. It now uses the helper with padjust, and unknown methods raise ValueError. The existing Holm doctest already showed the correct values and now actually passes. Also removed the commented-out 0-padded block.
- Fixed a regression that the NaN placeholders exposed: sidak() with an empty test family (every p-value NaN, e.g. a constant column) returned 0.0 because 1 - (1 - nan)**0 == 0, so an undefined correlation was starred ***. It now returns NaN like the other methods.

b297cc3 – Move rcorr family tests into test_correlation and strengthen them
- Moved the tests from the standalone tests/test_rcorr_family.py into TestCorrelation.test_rcorr
- The fixture now has injected correlations. With five independent null columns, many adjusted p-values clipped to exactly 1.0, so 10 of the 25 parametrized cases passed even with the bug present.
- Stars are now checked positively against stars derived from independently adjusted p-values, instead of asserting "no stars" under a fixed null seed (which held by chance of the seed and would also pass under an all-NaN regression).
- Added the constant-column case for every method, plus small additions to test_sidak and test_ptests.

I'll let you do a review of these changes, and will merge the PR once I get your green light. Thanks

dnncha:
Thanks Raphael, I’ve reviewed both commits and I’m happy for you to merge. The shared helper is clearer, and good catches on `ptests` and the all-NaN Sidak case.

The four relevant test files pass locally, along with 43 extra reference and edge-case checks covering the helper and paired/unpaired `ptests`. The stronger correlation and star checks look good too. Thanks for taking this further.

raphaelvallat (approved):

